PR #8, opened 2018-10-01
update manifest files name.

aveshagarwal:


smarterclayton:
Please use labels and prow merge rather than force merging

PR #13, opened 2018-10-02
Fix verify generated bin data script

ravisantoshgudimetla:
Fix bin generated files, previously we were pointing to v3.10 assets which would have failed.

### DIFF
--- a/hack/update-generated-bindata.sh
+++ b/hack/update-generated-bindata.sh
@@ -3,7 +3,7 @@ set -e
 set -u
 set -o pipefail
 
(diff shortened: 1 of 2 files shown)


aveshagarwal:
@ravisantoshgudimetla can you fix the title and description as they are misleading? This PR is fixing the verify script only, not the generated bin data.

ravisantoshgudimetla:
@aveshagarwal Done

ravisantoshgudimetla:
@aveshagarwal anything else needed here? If not can you lgtm?

PR #21, opened 2018-10-16
Unify kube-scheduler-config configmap names

sttts:
The kube-scheduler-config configmap created by the render command will live in kube-system, next to the phase-2 daemonset.

The config of phase-3 is supposed to be created by the operator (called `cm.yaml` in the other control plane component operators).

### DIFF
--- a/bindata/bootkube/manifests/kube-system-kube-scheduler-config.yaml
+++ b/bindata/bootkube/manifests/kube-system-kube-scheduler-config.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: kube-scheduler-config
-  namespace: {{ .Namespace }}
+  namespace: kube-system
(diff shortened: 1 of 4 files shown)


aveshagarwal:
@sttts should we change https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/cmd/cluster-kube-scheduler-operator/render/render.go#L61 to kube-system then?

sttts:
> should we change

No. That namespace is the operator namespace (phase 3) which stays as it is. We had to switch the phase 2 namespace to kube-system because bootkube expects that. Hence, we are creating all configmaps+secrets in kube-system **and** in the operator namespace. Compare https://github.com/openshift/cluster-kube-controller-manager-operator/tree/master/bindata/bootkube/manifests.

aveshagarwal:
ok sounds good.

/lgtm

PR #26, opened 2018-10-30
render: temporary --disable-phase-2

sttts:
Counterpart of openshift/cluster-kube-apiserver-operator#91 and prerequisite for a https://github.com/openshift/cluster-kube-controller-manager-operator/pull/63 like change.


cmd/cluster-kube-scheduler-operator/render/render.go
```diff
@@ -103,7 +113,21 @@ func (r *renderOpts) Run() error {
...
 
-	return genericrender.WriteFiles(&r.generic, &renderConfig.FileConfig, renderConfig)
+	// TODO: remove after the transition in the installer to a phase-2 free bootstrapping
+	var filters []assets.FileInfoPredicate
+	if r.disablePhase2 {
+		filters = append(filters, func(info os.FileInfo) bool {
+			if strings.HasPrefix(info.Name(), "kube-system-") {
+				return false
+			}
+			if info.Name() == "daemonset-kube-scheduler-daemonset.yaml" {
```
aveshagarwal: i think name of the file here should match to `kube-scheduler-daemonset.yaml` as here https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/bindata/bootkube/manifests/kube-scheduler-daemonset.yaml ?

sttts: indeed, typo

sttts: fixed

mfojtik:
/lgtm

This is just temporary until we transition the installer.

PR #49, opened 2019-01-16
Makefile: switch to library-go build system

sttts:
As all other library-go based operators.


sttts:
@tnozicka I don't want to refactor e2e tests here. I want mainly to unblock the follow-up PRs replacing library-go.

sttts:
Remove travis.

sttts:
/retest

Scheduler seems to run normally.

Makefile
```diff
@@ -1,132 +1,42 @@
...
+all: build
+.PHONY: all
+
+# Codegen module needs setting these required variables
+CODEGEN_OUTPUT_PACKAGE :=github.com/openshift/cluster-kube-scheduler-operator/pkg/generated
+CODEGEN_API_PACKAGE :=github.com/openshift/cluster-kube-scheduler-operator/pkg/apis
+CODEGEN_GROUPS_VERSION :=kubescheduler:v1alpha1
+
+# Exclude e2e tests from unit testing
+GO_TEST_PACKAGES :=$(shell go list ./... | grep -v e2e)
```
tnozicka: this will ignore packages like `./pkg/library/see2enable`
how about `GO_TEST_PACKAGES=./pkg/... ./cmd/...`

sttts: I don't want to list them manually. But I forgot the `-w` for the grep. Fixing.

sttts: done

tnozicka: not really `/pkg/e2e/important` get's ignored as well. You'd need to grep for the string with a prefix or use what I suggested

sttts: that should be ignore I would say

Makefile
```diff
@@ -1,132 +1,42 @@
...
+all: build
+.PHONY: all
+
+# Codegen module needs setting these required variables
+CODEGEN_OUTPUT_PACKAGE :=github.com/openshift/cluster-kube-scheduler-operator/pkg/generated
+CODEGEN_API_PACKAGE :=github.com/openshift/cluster-kube-scheduler-operator/pkg/apis
+CODEGEN_GROUPS_VERSION :=kubescheduler:v1alpha1
+
+# Exclude e2e tests from unit testing
+GO_TEST_PACKAGES :=$(shell go list ./... | egrep -wv '^github.com/openshift/cluster-kube-scheduler-operator/test/e2e')
```
tnozicka: I don't think this this the best path to approach it, e.g. with adding `/test/integration` this grepping is gonna get nasty and it would be much cleaner to use just `GO_TEST_PACKAGES=./pkg/.. ./cmd/...`

tnozicka: nit: also technically correct would be to use `$(GO)` there

PR #51, opened 2019-01-17
Update OWNERS

sttts:


sttts:
/assign @ravisantoshgudimetla 

@ravisantoshgudimetla do you have anybody who could be a second approver? Would be good to have one, so we are not blocked when you are out.

ravisantoshgudimetla:
> @ravisantoshgudimetla do you have anybody who could be a second approver? Would be good to have one, so we are not blocked when you are out.

Makes sense @sttts, @sjenning would you like to be 2nd approver? 

@sttts - Since you have been involved in this project right from start, would be interested in becoming another approver?

sttts:
> @sttts - Since you have been involved in this project right from start, would be interested in becoming another approver?

I would be happy to. I would concentrate more on the plumbing, less on deep scheduler config topics.

PR #52, opened 2019-01-17
Add Seth/Stefan to owners

ravisantoshgudimetla:
Adding Stefan and Seth to owners.


sjenning:
lgtm
@ravisantoshgudimetla I guess you have to self lgtm here since you are currently the only owner.

ravisantoshgudimetla:
/retest

@sjenning - I think anyone in the org can lgtm, only approval is limited to approvers.

PR #50, opened 2019-01-16
Bump library-go

sttts:
Replaces https://github.com/openshift/cluster-kube-scheduler-operator/pull/45.

PR #54, opened 2019-01-25
add related objs

juanvallejo:
compiles - just need to test locally
Updates deps in order to bring in the new `.Status.RelatedObject` field from openshift/api

cc @soltysh @deads2k


ravisantoshgudimetla:
@juanvallejo - Does this bring in version too?

juanvallejo:
@ravisantoshgudimetla 
> Does this bring in version too?

It brings in `.Status.Versions` (operand versions)

juanvallejo:
Currently debugging this locally. Status controller appears to be crash looping here https://github.com/openshift/library-go/blob/master/pkg/operator/status/status_controller.go#L82 and never brings up the clusteroperator resource

ravisantoshgudimetla:
Great @juanvallejo - Thanks for this PR. I am wondering if we need https://github.com/openshift/cluster-kube-scheduler-operator/pull/50 if this PR is anyway bumping library-go.

/cc @joelsmith for visibility as he started looking into this.
/cc @sjenning

ravisantoshgudimetla:
> Currently debugging this locally. Status controller appears to be crash looping here https://github.com/openshift/library-go/blob/master/pkg/operator/status/status_controller.go#L82 and never brings up the clusteroperator resource

I have started adding making some changes on the same lines for adding a configmap in `GlobalUserSpecifiedConfigNamespace` but I can copy most of my changes from this PR if this passes. So, I will try this locally.

juanvallejo:
>  I am wondering if we need #50 if this PR is anyway bumping library-go.

Main goal behind this one is to bump openshift/api. Side effect does end up being that it brings in an even newer library-go.

juanvallejo:
I've opened https://github.com/openshift/cluster-kube-scheduler-operator/pull/55 to test this PR. It bumps library go to [commit https://github.com/openshift/library-go/commit/7043c4ecdf45532bf6237c886a1544b01fc46fdc](https://github.com/openshift/library-go/commit/7043c4ecdf45532bf6237c886a1544b01fc46fdc) which still adds support for .Status.RelatedObjects, without bringing in the most recent changes in `library-go/master`.

https://github.com/openshift/cluster-kube-scheduler-operator/pull/55 actually appears to be working fine. It adds the `.RelatedObjects` to the clusteroperator status, as expected, and no crash looping is observed

juanvallejo:
Quick update: was able to debug the issue. Updated this PR accordingly. It should be working as expected now.

juanvallejo:
Quick update: was able to debug the issue. Updated this PR accordingly. It should be working as expected now.

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -0,0 +1,31 @@
...
+	kubeClient kubernetes.Interface,
+	eventRecorder events.Recorder) (*resourcesynccontroller.ResourceSyncController, error) {
+
+	resourceSyncController := resourcesynccontroller.NewResourceSyncController(
+		operatorConfigClient,
+		kubeInformersForNamespaces,
+		kubeClient,
+		eventRecorder,
+	)
+	if err := resourceSyncController.SyncConfigMap(
```
juanvallejo: @ravisantoshgudimetla I'm assuming this is the wrong configmap for this operator. Are there any resources that this operator will need to sync?

juanvallejo: cc @deads2k ^

ravisantoshgudimetla: At this point of time, no, I am planning to add one soon.

juanvallejo: Okay. Will remove this block from the PR then. Will still leave the empty resourcesynccontroller though, to at least make adding those easier in the future

deads2k:
> @juanvallejo - Thanks for the PR, the latest changes are bringing in all the fields we need but I am curious, why versions are being shown as null.

The versions API updated to reflect the potential for 0, 1, or many versions.  You'll need to separately update that aspect.

ravisantoshgudimetla:
Got it @deads2k. Thank you..

/hold cancel

PR #57, opened 2019-02-02
Add configmap in openshift-config namespace and watch it

ravisantoshgudimetla:
At a high level, following is what this PR does:

- When cluster-admin creates a configmap with name policy-configmap in openshift-config namespace, the kube-scheduler operator creates a new configmap with  same name in openshift-kube-scheduler namespace. 
- It continuously watches for configmap in openshift-configmap namespace, when changes are made to configmap in openshift-configmap, the same changes are made to configmap in openshift-kube-scheduler operator.
- When configmap is deleted openshift-configmap namespace, the corresponding configmap is deleted and scheduler falls back to default provider.

/cc @sjenning


pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -78,17 +79,34 @@ func createTargetConfigReconciler_v311_00_to_latest(c TargetConfigReconciler, re
...
+	}
 	requiredConfigMap, _, err := resourcemerge.MergeConfigMap(configMap, "config.yaml", nil, defaultConfig, operatorConfig.Spec.ObservedConfig.Raw, operatorConfig.Spec.UnsupportedConfigOverrides.Raw)
 	if err != nil {
 		return nil, false, err
 	}
 	return resourceapply.ApplyConfigMap(client, recorder, requiredConfigMap)
 }
 
-func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *v1alpha1.KubeSchedulerOperatorConfig, imagePullSpec string) (*corev1.ConfigMap, bool, error) {
+func managePod_v311_00_to_latest(lister corev1listers.ConfigMapLister, client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *v1alpha1.KubeSchedulerOperatorConfig, imagePullSpec string) (*corev1.ConfigMap, bool, error) {
```
ravisantoshgudimetla: The lister argument is not needed. Will remove it.

bindata/v3.11.0/kube-scheduler/defaultconfig-postbootstrap-with-policy.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: componentconfig/v1alpha1
+kind: KubeSchedulerConfiguration
+clientConnection:
+  kubeconfig: /etc/kubernetes/static-pod-resources/secrets/scheduler-kubeconfig/kubeconfig
+algorithmSource:
```
gnufied: Shouldn't we installing default policies in some environments? For example - it is pointless to enable Cinder, Azure, GCE-PD predicates in Openshift-4.0 environment on top of AWS?

sjenning: it is pointless but doesn't hurt anything

gnufied: There is a slight performance improvement for sure if you disable the unneeded predicates. But reality is - we can't use same default config across all platforms fore sure. For example - I am looking to enable a Cinder specific predicate which is disabled by default in scheduler. 

I guess in this case - we will have to create a configmap with cinder predicate in `openshift-kube-scheduler`  namespace by default via this operator if deployed on Openstack and operator will merge user specified configmap with our own configmap? Or we could install Cinder specific predicate in all environments.

ravisantoshgudimetla: @gnufied 

- The configmap needs to be created to in `openshift-config` namespace not in `openshift-kube-scheduler` namespace. 
- If you don't create in configmap in `openshift-config` namespace, we would start using `Default provider` similar to our old behaviour.(without scheduler-policy.json in older releases)
- If you want to enable cinder specific predicate, you need to create a configmap with cinder predicate enabled in OpenShift config namespace. I will create a docs PR which talks about how  to use the policy-configmap in 4.0 once I am back from vacation.

gnufied: I understand how it works but I thought creating configmap in `openshift-config` was cluster-admin action? We want to ship Cinder predicate enabled by default in Openshift-4.0 installation on Openstack. This is not something admin should have to configure. 

So lets say - if we did create the configmap `policy-configmap` in `openshift-config` namespace - so I guess admin will have to edit the configmap to make any changes to scheduler policy?

ravisantoshgudimetla: I'd say creating configmap in openshift-config is the only way at this point of time. Usually it's done through UI as post-install operations. Let me check if there is a generic way for defining set of post-install operations in 4.0. If it exists, you can perhaps add the policy configmap there.

mburke5678: @ravisantoshgudimetla 
> If you don't create in configmap in `openshift-config` namespace, we would start using `Default provider` similar to our old behaviour.(without scheduler-policy.json in older releases)

What is the 'Default provider` bahavior?

ravisantoshgudimetla: It is similar to what we have currently in 3.11, https://docs.openshift.com/container-platform/3.11/admin_guide/scheduling/scheduler.html#default-predicates, https://docs.openshift.com/container-platform/3.11/admin_guide/scheduling/scheduler.html#default-priorities. In 4.0, we have additional priority available imagelocality..

PR #59, opened 2019-02-12
Generate CRD validations with kubebuilder

damemi:
Running `update-crds` will `go get` kubebuilder crd tool, generate the new CRD with validations based on Go types, and replace the current CRD file with the new one after some bash hack to get rid of buggy output from kubebuilder.

This depends on https://github.com/kubernetes-sigs/controller-tools/pull/142 for the new `singular` tag

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,3 +40,23 @@ e2e: test-unit
 clean:
 	$(RM) ./cluster-kube-scheduler-operator
 .PHONY: clean
+
+OUTPUT_CRD=kubescheduler_v1alpha1_kubescheduleroperatorconfig.yaml
+CRD_MANIFEST=0000_11_kube-scheduler-operator_01_config.crd.yaml
+update-crds:
+	go get sigs.k8s.io/controller-tools/cmd/crd
+	crd generate --domain operator.openshift.io --output-dir manifests/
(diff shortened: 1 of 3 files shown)


sjenning:
This component is actually under the Pod team, not Master.
/assign
Also, I don't see this change in the kube-apiserver-operator or kube-controller-manager operator.  Any reason there is only a PR for this operator?

damemi:
This PR is only in this repo right now because the CRD generation tool
requires some changes that are in progress to be able to work with
operators that vendor their go types from openshift/api (such as
apiserver-operator, for example). So this is kind of a proof of concept to
demonstrate what we can do while we work on those changes

sjenning:
/approve
just need to rebase and pass the tests

damemi:
/hold
didn't notice [this change](https://github.com/openshift/cluster-kube-scheduler-operator/pull/61) which put this repo in the same state as the others by vendoring the types in openshift/api. so this needs to be updated once the [change in controller tools](https://github.com/kubernetes-sigs/controller-tools/pull/144) merges (or we put together a hack script a la https://github.com/openshift/cluster-config-operator/pull/17)

damemi:
closing in favor of https://github.com/openshift/cluster-kube-scheduler-operator/pull/67

PR #60, opened 2019-02-15
Bump the library-go and bump kube to 1.12.5

mfojtik:
In reaction, force to use protobuf client in operator.

/cc @sttts 
/cc @ravisantoshgudimetla


pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -18,7 +18,8 @@ func NewResourceSyncController(
 	resourceSyncController := resourcesynccontroller.NewResourceSyncController(
 		operatorConfigClient,
 		kubeInformersForNamespaces,
-		kubeClient,
+		kubeClient.CoreV1(),
+		kubeClient.CoreV1(),
```
sttts: this is what?

mfojtik: switch to Getters, it is good change

deads2k:
known flakes.  We're going to be in here adjusting API.  I'm going for this.

Issue #53, opened 2019-01-17
Use rotated client cert to contact apiserver


sttts:
The kube-apiserver operator rotates the client CA and creates control plane client certs. The one for the scheduler must be wired up as well, compare https://github.com/openshift/cluster-kube-apiserver-operator/pull/205/files#diff-ef02f4dd56d4bd19d7b3e4913b9a57d5R260.

PR #62, opened 2019-02-18
Rotate kubeconfig

ravisantoshgudimetla:
Following the PR https://github.com/openshift/cluster-kube-controller-manager-operator/pull/154

Previously, we were having a single kubeconfig coming from installer for all control plane components. This PR ensures we are separating that kubeconfig in kube-scheduler operator and allowing individual operator to have more control on the kubeconfig.


ravisantoshgudimetla:
/hold 

I haven't tested this on a cluster yet, facing issues while spinning up AWS cluster. Will spin up and test locally.

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -27,5 +27,11 @@ func NewResourceSyncController(
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalUserSpecifiedConfigNamespace, Name: "policy-configmap"}); err != nil {
 		return nil, err
 	}
+	if err := resourceSyncController.SyncSecret(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "kube-scheduler-client-cert-key"},
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "kube-scheduler-client-cert-key"},
```
ravisantoshgudimetla: @sttts - This should solve the https://github.com/openshift/cluster-kube-scheduler-operator/issues/53 too. Am I right?

ravisantoshgudimetla:
I swear, I saw issues related to 
`leaderelection.go:270] error retrieving resource lock kube-system/kube-scheduler: endpoints "kube-scheduler" is forbidden: User "system:kube-scheduler" cannot get resource "endpoints" in API group "" in the namespace "kube-system"`

Not sure, what is causing everything to work fine now..

ravisantoshgudimetla:
@deads2k - Do you have any idea on why this was failing previously and now it's working fine? I was able to reproduce this locally on my cluster before but not now, I have logs for serial run in case you're interested when it failed.

Also, if don't have any other concerns, I will tag this PR with lgtm label.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -130,3 +137,21 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
 	configMap.Data["version"] = version.Get().String()
 	return resourceapply.ApplyConfigMap(client, recorder, configMap)
 }
+
+func manageServiceAccountCABundle(lister corev1listers.ConfigMapLister, client coreclientv1.ConfigMapsGetter, recorder events.Recorder) (*corev1.ConfigMap, bool, error) {
```
deads2k: This can now be reduced to a resourceSync against `operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "kube-apiserver-server-ca"`

deads2k:
just the one comment that should be fixed up.  lgtm otherwise

ravisantoshgudimetla:
self-tagging as discussed over slack.

PR #70, opened 2019-03-01
Switch to getting policy configmap name from api

ravisantoshgudimetla:
This helps in getting policy configmap name from api instead of using hardcoded `policy-configmap`.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -15,6 +15,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions"
 	operatorversionedclient "github.com/openshift/client-go/operator/clientset/versioned"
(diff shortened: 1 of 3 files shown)


sjenning:
fyi @deads2k

sjenning:
My way of thinking is that we get rid of the `policy-configmap` hardcoded name all together.  Either the customer creates a configmap in `openshift-config` and create a `Scheduling` resource that point at it, or they get the default config.

sjenning:
when https://github.com/openshift/cluster-kube-scheduler-operator/pull/71 lands, this will need rebase with s/Scheduling/Scheduler/ in the config type.

ravisantoshgudimetla:
> My way of thinking is that we get rid of the policy-configmap hardcoded name all together. Either the customer creates a configmap in openshift-config and creates a Scheduling resource that point at it in the policy spec field , or they get the default config.

I agree, I made the change, the only hard-coding that exists as of now is the `policy-configmap` that gets created within `openshift-kube-scheduler` namespace. I am having it because we have that hardcoding in the bindata/post-bootstrap files. But the user specified configmap will be still available in `openshift-config` namespace, while copying the file, we are changing the name to `policy-configmap` because of bindata hardcoding. I will remove this when we move to config observer based mechanism.

PR #72, opened 2019-03-03
Get feature gates and using them while starting kube-scheduler

ravisantoshgudimetla:
I was initially thinking of using config observer pattern to list feature gates and update kube-scheduler but later realized that we are updating any config related to kube-scheduler here but just updating one of the flags, so I have updated the target_config_reconciler.go. I haven't tested this locally.

### DIFF
--- a/pkg/operator/target_config_reconciler_v311_00.go
+++ b/pkg/operator/target_config_reconciler_v311_00.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/klog"
 	"reflect"
 
+	"github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -53,7 +54,7 @@ func createTargetConfigReconciler_v311_00_to_latest(c TargetConfigReconciler, re
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/serviceaccount-ca", err))
 	}
-	_, _, err = managePod_v311_00_to_latest(c.kubeClient.CoreV1(), recorder, operatorConfig, c.targetImagePullSpec)
+	_, _, err = managePod_v311_00_to_latest(c.kubeClient.CoreV1(), recorder, operatorConfig, c.targetImagePullSpec, c.featureGateLister)
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/kube-scheduler-pod", err))
 	}
@@ -132,11 +133,25 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
 	return resourceapply.ApplyConfigMap(client, recorder, requiredConfigMap)
 }
 
-func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorv1.KubeScheduler, imagePullSpec string) (*corev1.ConfigMap, bool, error) {
+func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorv1.KubeScheduler, imagePullSpec string, featureGateLister configlistersv1.FeatureGateLister) (*corev1.ConfigMap, bool, error) {
 	required := resourceread.ReadPodV1OrDie(v311_00_assets.MustAsset("v3.11.0/kube-scheduler/pod.yaml"))
 	if len(imagePullSpec) > 0 {
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
+	// check for feature gates from feature lister.
+	featureGates, err := checkForFeatureGates(featureGateLister)
+	if err != nil || len(featureGates) == 0 {
+		// Default behaviour
+		klog.Infof("Error while getting feature gates or there are no featuregates for given feature, so defaulting to empty feature gates")
+		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("--feature-gates=%v", ""))
+	} else {
+		allFeatureGates := ""
+		for featureGate, status := range featureGates {
+			allFeatureGates = allFeatureGates + "," + fmt.Sprintf("%v=%v", featureGate, status)
+		}
+		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("--feature-gates=%v", allFeatureGates))
+	}
+
 	switch operatorConfig.Spec.LogLevel {
 	case operatorv1.Normal:
 		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", 2))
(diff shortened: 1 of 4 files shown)


pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -137,6 +147,42 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
...
+	featureGateListConfig, err := featureGateLister.Get("cluster")
+	var featureGates = make(map[string]bool)
+	if apierrors.IsNotFound(err) {
+		glog.Infof("features.config.openshift.io/cluster: not found, so return empty feature gates")
+		return nil, err
+	}
+	var enabledFeatureSets, disabledFeatureSets []string
+	currentFeatureSetConfig := featureGateListConfig.Spec.FeatureSet
+	switch {
+	case currentFeatureSetConfig == "" || currentFeatureSetConfig == v1.Default:
```
ravisantoshgudimetla: @rphillips - Is this how feature gates are intended to be used by caller or did I miss something?

rphillips: v1.Default === "" so you can just pull the FeatureSet out with:

featureSet, ok := v1.FeatureSets[currentFeatureSetConfig]

No need for the switch.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -137,6 +147,42 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
 	return resourceapply.ApplyConfigMap(client, recorder, configMap)
 }
 
+func checkForFeatureGates(featureGateLister configlistersv1.FeaturesLister) (map[string]bool, error) {
+	featureGateListConfig, err := featureGateLister.Get("cluster")
```
rphillips: do we know the cluster wide featureGate will be named cluster?

ravisantoshgudimetla: Nope, I was using same name that we used for other CRs.

pkg/operator/target_config_reconciler_v311_00.go
ravisantoshgudimetla: Ya, my question was more on the lines of the usage of `currentFeatureSetConfig` but I think I am doing it right. Thank you :). Will remove switch..

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -151,6 +162,34 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
...
+	var featureGates = make(map[string]bool)
+	if apierrors.IsNotFound(err) {
+		glog.Infof("features.config.openshift.io/cluster: not found, so return empty feature gates")
+		return nil, err
+	}
+	var enabledFeatureSets, disabledFeatureSets []string
+	currentFeatureSetConfig := featureGateListConfig.Spec.FeatureSet
```
rphillips: I think it's a little cleaner passing in the featureSet into the generateFeatureGates function

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -126,11 +127,21 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
...
+func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorv1.KubeScheduler, imagePullSpec string, featureGateLister configlistersv1.FeaturesLister) (*corev1.ConfigMap, bool, error) {
 	required := resourceread.ReadPodV1OrDie(v311_00_assets.MustAsset("v3.11.0/kube-scheduler/pod.yaml"))
 	if len(imagePullSpec) > 0 {
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
+	// check for feature gates from feature lister.
+	featureGates, err := checkForFeatureGates(featureGateLister)
+	if err != nil {
+		// Default behaviour
+		glog.Infof("Error while getting feature gates, so defaulting to empty feature gates")
```
rphillips: should log the error here.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -151,6 +162,34 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
 	return resourceapply.ApplyConfigMap(client, recorder, configMap)
 }
 
+func checkForFeatureGates(featureGateLister configlistersv1.FeaturesLister) (map[string]bool, error) {
+	featureGateListConfig, err := featureGateLister.Get("cluster")
+	var featureGates = make(map[string]bool)
+	if apierrors.IsNotFound(err) {
```
rphillips: need to check for err that isn't a not found error as well

pkg/operator/target_config_reconciler_v311_00.go
ravisantoshgudimetla: I prefer to either return error or log error but not both. I forgot to log it at the source, I did it there I think no need to do it here.

pkg/operator/target_config_reconciler_v311_00.go
ravisantoshgudimetla: Good catch, made the change. Thank you :)

eparis:
https://github.com/openshift/api/pull/278

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -132,11 +133,25 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
...
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
+	// check for feature gates from feature lister.
+	featureGates, err := checkForFeatureGates(featureGateLister)
+	if err != nil || len(featureGates) == 0 {
+		// Default behaviour
+		klog.Infof("Error while getting feature gates or there are no featuregates for given feature, so defaulting to empty feature gates")
+		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("--feature-gates=%v", ""))
+	} else {
+		allFeatureGates := ""
```
ravisantoshgudimetla: I have added this recently.

rphillips: featuregates need a true or false to designate whether they are enabled or disabled.

ravisantoshgudimetla: Sorry, can you please elaborate? I am constructing a string with "," separated values here so that I can pass this onto `--feature-gates` flag to kube-scheduler.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -157,6 +172,34 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
 	return resourceapply.ApplyConfigMap(client, recorder, configMap)
 }
 
+func checkForFeatureGates(featureGateLister configlistersv1.FeatureGateLister) (map[string]bool, error) {
+	featureGateListConfig, err := featureGateLister.Get("cluster")
+	var featureGates = make(map[string]bool)
+	if err != nil {
```
rphillips: Need to pull in the default FeatureSet if the cluster FeatureGate cannot be found.

pkg/operator/target_config_reconciler_v311_00.go
rphillips: sorry. I'll comment in the right spot.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -132,11 +133,25 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
...
 }
 
-func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorv1.KubeScheduler, imagePullSpec string) (*corev1.ConfigMap, bool, error) {
+func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, operatorConfig *operatorv1.KubeScheduler, imagePullSpec string, featureGateLister configlistersv1.FeatureGateLister) (*corev1.ConfigMap, bool, error) {
 	required := resourceread.ReadPodV1OrDie(v311_00_assets.MustAsset("v3.11.0/kube-scheduler/pod.yaml"))
 	if len(imagePullSpec) > 0 {
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
+	// check for feature gates from feature lister.
+	featureGates, err := checkForFeatureGates(featureGateLister)
```
rphillips: if we are pulling in the default featuregates from openshift/api, then this function should never error.

pkg/operator/target_config_reconciler_v311_00.go
ravisantoshgudimetla: Done. PTAL.

pkg/operator/target_config_reconciler_v311_00.go
ravisantoshgudimetla: Right, the latest commit has no error as 2nd return value.

ravisantoshgudimetla:
Seems `kube-apiserver` is erroring out.

`  "message": "StaticPodsFailing: pods \"kube-apiserver-ip-10-0-136-241.ec2.internal\" not found\nStaticPodsFailing: pods \"kube-apiserver-ip-10-0-173-123.ec2.internal\" not found",
                        "reason": "StaticPodsFailingError",`

PR #75, opened 2019-03-20
Bump(*) dependencies

ravisantoshgudimetla:
Bump to get 

https://github.com/openshift/client-go/pull/97/

/cc @sjenning 

FYI, we are getting what we need, https://github.com/ravisantoshgudimetla/cluster-kube-scheduler-operator/blob/389ff01825f9014c5d20fef29b73efc2e8f47381/vendor/github.com/openshift/client-go/config/clientset/versioned/typed/config/v1/featuregate.go


ravisantoshgudimetla:
@damemi - We need to check if we need changes you mentioned in https://github.com/openshift/cluster-kube-scheduler-operator/pull/74 and if this bump is going to fail without the changes in https://github.com/openshift/cluster-kube-scheduler-operator/pull/74/commits/8416646582aca146cf6ba018b5158b0d30956da4

damemi:
@ravisantoshgudimetla without that change I couldn't get this to compile, as you can see from the CI errors. But I'm also still unable to get CI to pass in #74 so there could be some other problems, if you want this PR to overtake that one I'm not opposed

sjenning:
@ravisantoshgudimetla if still needed, this will need rebase on https://github.com/openshift/cluster-kube-scheduler-operator/pull/74

ravisantoshgudimetla:
@sjenning - I think, this supersedes the changes in #74, so I need to get this working anyways.

PR #76, opened 2019-03-21
Policy configmap working with kube-scheduler

ravisantoshgudimetla:
This PR adds

- RBAC for `system:kube-scheduler` user to access configmaps in `openshift-kube-scheduler` namespace. This was not an issue before #62 merged but after that we have to add following roles/rolebindings.

/cc @sjenning 

We are able to pick up configmap changes now.

`I0321 23:06:23.187561       1 factory.go:1118] Creating scheduler from configuration: {{ } [{PodFitsHostPorts <nil>} {PodFitsResources <nil>} {NoDiskConflict <nil>} {NoVolumeZoneConflict <nil>} {MatchNodeSelector <nil>} {HostName <nil>}] [{LeastRequestedPriority 1 <nil>} {BalancedResourceAllocation 1 <nil>} {ServiceSpreadingPriority 1 <nil>} {EqualPriority 1 <nil>}] [] 0 false}
`
We can see above line in kube-scheduler logs.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	configv1 "github.com/openshift/api/config/v1"
+
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions"
@@ -79,6 +80,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	resourceSyncController, err := resourcesynccontroller.NewResourceSyncController(
 		operatorClient,
 		kubeInformersForNamespaces,
+		configInformers,
 		kubeClient,
 		ctx.EventRecorder,
 	)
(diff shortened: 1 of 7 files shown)


pkg/operator/starter.go
```diff
@@ -146,6 +148,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	kubeInformersClusterScoped.Start(ctx.Done())
 	kubeInformersNamespace.Start(ctx.Done())
 	kubeInformersForNamespaces.Start(ctx.Done())
+	configInformers.Start(ctx.Done())
```
ravisantoshgudimetla: This was something I missed last time.

pkg/operator/v311_00_assets/bindata.go
```diff
@@ -344,6 +346,66 @@ func v3110KubeSchedulerPodYaml() (*asset, error) {
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: openshift-kube-scheduler
+  name: system:openshift:sa-listing-configmaps
+roleRef:
+  kind: Role
+  name: system:openshift:sa-listing-configmaps
+subjects:
+- kind: ServiceAccount
```
ravisantoshgudimetla: The generated data is wrong. I missed two commits which is causing this PR to fail.

Note to myself: Add verify-generated bindata..

ravisantoshgudimetla: /cc @sjenning

PR #79, opened 2019-03-26
Update bindata

ravisantoshgudimetla:
I missed two commits while pushing changes related to #76 which is causing to generate a different rolebindings.

/cc @sjenning.

will add an e2e related to this subsequent PRs.


sjenning:
/retest
verify should work now https://github.com/openshift/release/pull/3249

sttts:
fyi, with another library-go bump bindata will always be checked in CI.

PR #77, opened 2019-03-26
Set cert-dir to prepare for 1.13

sttts:
Due to https://github.com/kubernetes/kubernetes/pull/69884.

PR #80, opened 2019-03-26
force scheduler insecure

deads2k:
this will help land the rebase.  The secure option fails to bootstrap.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
@@ -15,7 +15,7 @@ spec:
     imagePullPolicy: {{ .ImagePullPolicy }}
     command: ["/bin/bash", "-c"]
     args:
-    - exec hyperkube kube-scheduler --kubeconfig=/etc/kubernetes/secrets/kubeconfig --leader-elect=true
+    - exec hyperkube kube-scheduler --kubeconfig=/etc/kubernetes/secrets/kubeconfig --leader-elect=true --secure-port=0 --port=10251
     volumeMounts:
     - mountPath: /etc/kubernetes/secrets
       name: secrets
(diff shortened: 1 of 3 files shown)


sjenning:
fyi @smarterclayton @derekwaynecarr 
why do we need to do this?
/hold

deads2k:
> fyi @smarterclayton @derekwaynecarr
> why do we need to do this?
> /hold

The rebase is blocked on this because in 1.13 our dynamic reload patch interacts badly with the cert generation change that no longer serializes to disk.  This happens because the cert for this component is missing. This combination results in a case where kube-scheduler panics and dies.  Choices we have to land the rebase
1. operator fixes the missing cert by introducing a rotated cert to manage this
2. operator fixes the missing cert by conditionally enabling secure serving based on availability of a generated serving cert.
3. remove our dynamic reloading
4. change a dynamic reloading path with an untestable scenario in 1.12 that attempts to allow illegal conditions and may still fail in the rebase
5. hard code the rebase to ignore the operator flags and start insecure regardless
6. revert upstream pulls that changed cert serialization
7.  Or what this pull does, simply serve insecurely for a week to land the rebase.

We tried to work around this in another pull, but that pull wasn't able to bootstrap the server.

sjenning:
@deads2k dynamic reload patch? what is this? link to source?

So there will need to be a follow-on to this to reenable secure connections after the rebase?

deads2k:
> So there will need to be a follow-on to this to reenable secure connections after the rebase?

yes



> @deads2k dynamic reload patch? what is this? link to source?

It's spread across 6-10 pulls.  https://github.com/openshift/kubernetes/commit/00438e8cb90ff3302de8006ce2f7f78663dfab78 is most of it in 1.13, but it's slightly different in 1.12

soltysh:
@sjenning we tried to land #77 but `--cert-dir` is not a recognized flag by the `kube-scheduler` although it's wired there :man_shrugging: anyway, let's try to land this to unstuck the rebase.

deads2k:
ok, it's failin because https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/branch-ci-openshift-origin-master-images/665

deads2k:
I'm pretty sure this worked, but I'll wait for a cleaner run

ravisantoshgudimetla:
Yay, finally all tests passed :)

PR #87, opened 2019-04-03
Upgrade fix

deads2k:
This switches us to use and env var


ravisantoshgudimetla:
Thanks @deads2k 

/lgtm

PR #88, opened 2019-04-04
Disable insecure port

sttts:
This is blocked by https://github.com/openshift/cluster-monitoring-operator/blob/172beb3e0d2cbf1f79bf61ba9c3f5bea373b4125/assets/prometheus-k8s/service-monitor-kube-scheduler.yaml#L15 pointing to the secure port.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
@@ -15,7 +15,7 @@ spec:
     imagePullPolicy: {{ .ImagePullPolicy }}
     command: ["/bin/bash", "-c"]
     args:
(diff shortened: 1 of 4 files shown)


s-urbaniak:
just a small request from my side, I enabled a tls based servicemonitor locally, but the kube-scheduler endpoints seem to point to the wrong port:
```
$ kubectl -n openshift-kube-scheduler get ep scheduler -o yaml
apiVersion: v1
kind: Endpoints
metadata:
...
  ports:
  - name: https
    port: 8443
    protocol: TCP
```
I believe these should rather be `10259`. Is it possible to fix this in conjunction with this change?

s-urbaniak:
/lgtm
I guess we'll have some chicken and egg integration woes here in conjunction with cluster monitoring, but this should be resolved once https://github.com/openshift/cluster-monitoring-operator/pull/308 and this PR both are merged.

PR #93, opened 2019-04-09
Bump with switch to klog

ravisantoshgudimetla:
@sjenning - Klog switch along with bump to remove glog.


eparis:
https://github.com/openshift/api/pull/278

ravisantoshgudimetla:
@sjenning - Adding LGTM assuming, it passes tests..

PR #94, opened 2019-04-10
prevent panic in logging message

deads2k:
I've observed a panic from this line for logging that stopped my cluster from starting.

### DIFF
--- a/pkg/operator/target_config_reconciler_v311_00.go
+++ b/pkg/operator/target_config_reconciler_v311_00.go
@@ -122,7 +122,14 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
 			defaultConfig = v311_00_assets.MustAsset("v3.11.0/kube-scheduler/defaultconfig-postbootstrap.yaml")
 		}
 	} else {
-		klog.Infof("Error while getting scheduler type %v and using default algorithm provider in kubernetes scheduler", err.Error())
+		msg := "unknown"
+		switch {
+		case err != nil:
+			msg = err.Error()
+		case len(observedpolicyConfigMap.Spec.Policy.Name) == 0:
+			msg = "missing policy"
+		}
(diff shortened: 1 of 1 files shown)


deads2k:
blocks config.openshift.io resource creation.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -122,7 +122,14 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
 			defaultConfig = v311_00_assets.MustAsset("v3.11.0/kube-scheduler/defaultconfig-postbootstrap.yaml")
 		}
 	} else {
-		klog.Infof("Error while getting scheduler type %v and using default algorithm provider in kubernetes scheduler", err.Error())
+		msg := "unknown"
+		switch {
```
ravisantoshgudimetla: BTW, we're already checking for error here https://github.com/openshift/cluster-kube-scheduler-operator/pull/94/files#diff-c65577daa1eb9509fc39b6af03a81eeeR97, so perhaps removing `err=nil ` in https://github.com/openshift/cluster-kube-scheduler-operator/pull/94/files#diff-c65577daa1eb9509fc39b6af03a81eeeR101 is ok and we can remove the switch altogether.

deads2k:
unsticking pulls in other repos.  merging

pkg/operator/target_config_reconciler_v311_00.go
deads2k: > BTW, we're already checking for error here https://github.com/openshift/cluster-kube-scheduler-operator/pull/94/files#diff-c65577daa1eb9509fc39b6af03a81eeeR97, so perhaps removing `err=nil ` in https://github.com/openshift/cluster-kube-scheduler-operator/pull/94/files#diff-c65577daa1eb9509fc39b6af03a81eeeR101 is ok and we can remove the switch altogether.

after https://github.com/openshift/cluster-config-operator/pull/15 merges, feel free to refactor

PR #98, opened 2019-04-12
library-go

deads2k:
brings in:

1. cleaned up api
1. status momentum to limit failing blips
1. operator logging levels


deads2k:
API is part of beta.  tagging.

PR #110, opened 2019-04-24
 Change one of the ports to be checked to 10259 

ravisantoshgudimetla:
Will close #108 once this passes


ravisantoshgudimetla:
@sjenning adding lgtm

PR #116, opened 2019-04-26
bump(*): library-go to include revision controller deletion recovery

damemi:
Includes changes from https://github.com/openshift/library-go/pull/368 to recover from operator resource deletion


sjenning:
a recent bump merged in https://github.com/openshift/cluster-kube-scheduler-operator/pull/119

PR #119, opened 2019-04-29
change key for cluster-operator version

deads2k:
This allows us to separate changes to the operator level from changes in the overall level.  Static pods can consume information via side-cars and so we need to know when the operator was updated and when the consistent level was achieved.  This is particular important for upgrades and this should enhance upgrade stability.


deads2k:
waiting on https://github.com/openshift/cluster-kube-scheduler-operator/pull/118

deads2k:
terraform  We really need to see this pass an upgrade

/retest

PR #124, opened 2019-05-02
Bug 1705649 : Cluster with halted master did not reschedule operators after 5m of being down

ravisantoshgudimetla:
As of now, because of infinite tolerations against all the possible taints, we are seeing that operators are not getting evicted from nodes that have `NoExecute` taint on them. This PR tightens the conditions around which can be pods can be scheduled/evicted. The downside is there is a very good chance that pods would be evicted from nodes that have certain conditions like `disk-pressure`, `memory-pressure`, taints added by other controllers(operators) etc. So, please make sure that this change is ok with your operator/operand before merging this PR.

/cc @sjenning @smarterclayton

### DIFF
--- a/manifests/0000_25_kube-scheduler-operator_06_deployment.yaml
+++ b/manifests/0000_25_kube-scheduler-operator_06_deployment.yaml
@@ -24,7 +24,7 @@ spec:
         command: ["cluster-kube-scheduler-operator", "operator"]
         args:
         - "--config=/var/run/configmaps/config/config.yaml"
-        - "-v=4"
+        - "-v=2"
         resources:
           requests:
             memory: 50Mi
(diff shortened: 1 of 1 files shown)


bindata/v3.11.0/kube-scheduler/pod.yaml
```diff
@@ -58,7 +58,17 @@ spec:
...
   priorityClassName: system-node-critical
   tolerations:
-  - operator: "Exists"
+   - key: node-role.kubernetes.io/master  # Just tolerate NoSchedule taint on master node. If there are other conditions like disk-pressure etc, let's not schedule the control-plane pods onto that node.
+     operator: Exists
+     effect: "NoSchedule"
+   - key: "node.kubernetes.io/unreachable"
+     operator: "Exists"
+     effect: "NoExecute"
+     tolerationSeconds: 120 # Evict pods within 2 mins.
```
ravisantoshgudimetla: This should help in evicting within 2 min @smarterclayton

manifests/0000_25_kube-scheduler-operator_06_deployment.yaml
```diff
@@ -24,7 +24,7 @@ spec:
         command: ["cluster-kube-scheduler-operator", "operator"]
         args:
         - "--config=/var/run/configmaps/config/config.yaml"
-        - "-v=4"
+        - "-v=2"
```
ravisantoshgudimetla: @sjenning - I included your change from #120

bindata/v3.11.0/kube-scheduler/pod.yaml
smarterclayton: This is good.  How do we get this to all other operators (once we're clear here that this works the way we expect)?

ravisantoshgudimetla: I tested this manually on my cluster. @sjenning and I plan to create PRs against all existing components.

PR #126, opened 2019-05-03
static-pod: fail liveness when secrets/kube-scheduler-client-cert-key changes

sttts:
We have to dynamically reload client certs in kas, kcm and ks in order to allow the cert recovery command https://github.com/openshift/cluster-kube-apiserver-operator/pull/444 to recover a broken cluster.

This is a quick and dirty way to archieve this. We plan to replace it with https://github.com/openshift/library-go/pull/392 when it is ready.

### DIFF
--- a/bindata/v3.11.0/kube-scheduler/pod.yaml
+++ b/bindata/v3.11.0/kube-scheduler/pod.yaml
@@ -42,12 +42,16 @@ spec:
     - mountPath: /etc/kubernetes/static-pod-resources
       name: resource-dir
     livenessProbe:
-      httpGet:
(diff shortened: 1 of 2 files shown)


sjenning:
Can we get descriptions on these PRs? :disappointed: I have no idea why we need to do this.  `exec` probes are awful too.

sttts:
@sjenning added the description. We know this is ugly. But we need a way to do that now. Compare the sidecar solution for the medium term solution.

ravisantoshgudimetla:
@sttts - Do I need to bump vendor to get this change from library-go?

PR #130, opened 2019-05-07
move v3.11.0 to v4.1.0

mfojtik:


mfojtik:
@ravisantoshgudimetla @sjenning 

This can't merge because I think you hardcoded this into your CI job definition:

```
make: *** No rule to make target `verify-bindata-v3.11.0'.  Stop.
```

We can override this check and land this and you fix your CI afterwards?

ravisantoshgudimetla:
> make: *** No rule to make target `verify-bindata-v3.11.0'.  Stop.

I am  fine with it. But should we have versions in the code? I think, for every release we might have to update it.

ravisantoshgudimetla:
> /override ci/prow/verify

Let me try that.

/override ci/prow/verify

ravisantoshgudimetla:
> @ravisantoshgudimetla: ravisantoshgudimetla unauthorized: /override is restricted to repo administrators

 Ohh no, perhaps @deads2k ?

soltysh:
Done in #170 
/close

PR #134, opened 2019-05-10
Vendor bump to library-go fixes

ravisantoshgudimetla:
Bump to get library-go dependencies.

/cc @sjenning

PR #135, opened 2019-05-11
Remove unnecessary rules

ravisantoshgudimetla:
Once https://github.com/openshift/cluster-kube-scheduler-operator/pull/134 merges, we can remove the role. Let's hope this passes tests :+1: 

/cc @sjenning


ravisantoshgudimetla:
upgrade passed

@sjenning if you're fine, can you lgtm this PR?

PR #136, opened 2019-05-29
Get policy configmap using config observer pattern instead of relying on static config file

ravisantoshgudimetla:
Initial pass on getting/updating policyconfigmap using config observer pattern

- [x]  Still need to test this locally
- [x] Remove the code in target config reconciler

/cc @sjenning


pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -0,0 +1,83 @@
...
+
+// observeSchedulerConfig lists the scheduler configuration and updates the name of the configmap that we want scheduler
+// to use as policy config
+func ObserveSchedulerConfig(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+	policyConfigMapRootPath := []string{"algorithmSource", "policy", "configMap"}
+	policyConfigMapNamePath := append(policyConfigMapRootPath, "name")
+	policyConfigMapNamespacePath := append(policyConfigMapRootPath, "namespace")
```
deads2k: I think this value should always be pinned.  I'm pretty sure that's what I'm going to read below, but please comment the intent

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -0,0 +1,83 @@
...
+		klog.Warningf("schedulers.config.openshift.io/cluster: not found")
+		return observedConfig, errs
+	}
+	if err != nil {
+		errs = append(errs, err)
+		return prevObservedConfig, errs
+	}
+	configMapName := schedulerConfig.Spec.Policy.Name
+
+	if len(configMapName) == 0 {
```
deads2k: if the value is cleared, we want to stop sync-ing, correct?  Check the doc, but I think you can pass an empty source to clear a target.

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -0,0 +1,83 @@
...
+		return observedConfig, errs
+	}
+
+	err = listers.ResourceSyncer().SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{
+			Namespace: operatorclient.TargetNamespace,
```
deads2k: you should always place this in the same spot in the same spot in the targetnamespace so you never have a collision.

deads2k:
You also need to react to changes to the configmap itself don't you?  Does the scheduler auto-detect them?  If not, you can look at the openshift-apiserver-operator which watches for changes to configmap and secret content to kick a new deployment.

pkg/operator/configobservation/scheduler/observe_scheduler.go
ravisantoshgudimetla: Isn't `observedConfig` empty here?

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -0,0 +1,83 @@
...
+	schedulerConfig, err := listers.SchedulerLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("schedulers.config.openshift.io/cluster: not found")
+		return observedConfig, errs
+	}
+	if err != nil {
+		errs = append(errs, err)
+		return prevObservedConfig, errs
+	}
+	configMapName := schedulerConfig.Spec.Policy.Name
```
ravisantoshgudimetla: Note to myself: Check if Spec is nil and Policy is empty..

pkg/operator/configobservation/scheduler/observe_scheduler.go
ravisantoshgudimetla: Can you elaborate? What do you mean `you should`? Isn't it always going to put it in the same spot?

pkg/operator/configobservation/scheduler/observe_scheduler.go
ravisantoshgudimetla: Sure. I will :)

ravisantoshgudimetla:
> You also need to react to changes to the configmap itself don't you? Does the scheduler auto-detect them? If not, you can look at the openshift-apiserver-operator which watches for changes to configmap and secret content to kick a new deployment.

When we make changes to configmap, we should make changes to the configmap in `openshift-config` namespace and resource sync controller would create a new version of it because it, it should restart Scheduler. Wouldn't that be case? I tested this scenario manually though and I am able to see a new version and scheduler restarting.

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if len(configMapName) == 0 {
+		return observedConfig, errs
+	}
+
+	err = listers.ResourceSyncer().SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{
```
sanchezl: You'll want this source `ResourceLocation` to be a zero object when `len(schedulerConfig.Spec.Policy.Name) == 0` so that a previously copied config maps (if any) is cleaned up.

ravisantoshgudimetla: Wouldn't we return early with observerConfig here(https://github.com/openshift/cluster-kube-scheduler-operator/pull/136/files#diff-759df316c2ba817633e0aaf1b6af0d8eR55) which has no `policyConfigMap`?

sanchezl: I just realized David said the same thing above. The idea is to not leave an unused config map in the target namespace when no longer needed.

ravisantoshgudimetla: Wait, shouldn't the deletion of configmap in `openshift-config` namespace trigger the deletion of this configmap?

ravisantoshgudimetla:
/retest

@sanchezl and @deads2k - any other comments

ravisantoshgudimetla:
``openshift-tests [sig-scheduling] SchedulerPreemption [Serial] validates basic preemption works [Suite:openshift/conformance/serial] [Suite:k8s] 1m27s``

@damemi - Same bug that you're working on.

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -0,0 +1,95 @@
...
+		if err := unstructured.SetNestedField(prevObservedConfig, currentPolicyConfigMapNamespace, policyConfigMapNamespacePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+	schedulerConfig, err := listers.SchedulerLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("schedulers.config.openshift.io/cluster: not found")
+		return observedConfig, errs
```
sanchezl: Delete (Sync w/ src==nil)  the configmap before returning.

ravisantoshgudimetla: This won't happen in the real clusters though. We'd recreate the `cluster` CR in the cluster config operator, if it doesn't exist

ravisantoshgudimetla: Done. PTAL

ravisantoshgudimetla:
Finally this merged.

@damemi - Do you want to cherrypick this for your bug in 4.1 branch or further debug?

PR #138, opened 2019-05-30
Bump kube to 1.14

ravisantoshgudimetla:
/cc @sjenning


ravisantoshgudimetla:
May 30 21:32:39.371: INFO: cluster upgrade is failing: Could not update oauthclient "console" (224 of 372): the server is down or not responding

Seems like unrelated issue

/retest

ravisantoshgudimetla:
The upgrade ran successfully twice. Tagging it.

PR #140, opened 2019-06-05
Fix operator status

ravisantoshgudimetla:
Fix operator status, so that we don't enter into permanent degrade status.

/cc @deads2k @sjenning


ravisantoshgudimetla:
xref: https://bugzilla.redhat.com/show_bug.cgi?id=1717618

ravisantoshgudimetla:
`level=error msg="  on ../tmp/openshift-install-346328721/vpc/master-elb.tf line 1, in resource \"aws_lb\" \"api_internal\":"

Seems like elb issue.

/retest

PR #148, opened 2019-06-21
updates dependencies

p0lyn0mial:
updates dependencies for `cluster-kube-scheduler-operator`


p0lyn0mial:
I had to run `make update-codegen-crds`  to update https://github.com/openshift/cluster-kube-scheduler-operator/pull/148/files#diff-b5af6b34980987785fc36ff2d374f987L56 because https://github.com/openshift/cluster-kube-scheduler-operator/pull/148/files#diff-2e6ec1d34cd474b4fc2ca77e12e53aa1R26 was updated

@mfojtik does this look suspicious to you? I think it wasn't intended to remove a validation pattern from the openapi spec.

p0lyn0mial:
> I had to run make update-codegen-crds to update https://github.com/openshift/cluster-kube-scheduler-operator/pull/148/files#diff-b5af6b34980987785fc36ff2d374f987L56 because https://github.com/openshift/cluster-kube-scheduler-operator/pull/148/files#diff-2e6ec1d34cd474b4fc2ca77e12e53aa1R26 was updated

> @mfojtik does this look suspicious to you? I think it wasn't intended to remove a validation pattern from the openapi spec.


@ironcladlou could you have a look? was this intended?

p0lyn0mial:
It seems like the current version of the generator (v0.1) doesn't know how to handle cases when a tag is attached to the type and not to the field. The issue has been fixed in v0.2 of the generator. I added a patch that adds the pattern until we switch to a new version. 

/hold cancel

@mfojtik @damemi thanks for your input
@mfojtik @damemi PTAL

damemi:
@p0lyn0mial I think you'll need to run `make update-codegen` to actually apply the patch to the crd if you didn't already

p0lyn0mial:
> @p0lyn0mial I think you'll need to run make update-codegen to actually apply the patch to the crd if you didn't already

all good, I did `make update-codegen-crds` :)

PR #149, opened 2019-06-27
manifests: add ServiceMonitor for kube-scheduler

paulfantom:
Since each team should own their monitoring configuration we are removing ServiceMonitor manifests from cluster-monitoring-operator.

Related to openshift/cluster-monitoring-operator#391

/cc @brancz @sttts


paulfantom:
Moved adding ServiceMonitor into bindata.

PR #152, opened 2019-07-11
Use new crd-schema-gen in Makefile

damemi:
This updates the crd generator in the Makefile to the new version, which also fixes the lack of top-level description in the validation/oc explain.


sttts:
lgtm
/approve
/hold
/assign @ravisantoshgudimetla 
for final lgtm.

PR #153, opened 2019-07-16
move kube-scheduler servicemonitor to manifests

paulfantom:
Move incorrectly placed ServiceMonitor manifest from `bindata/` into `manifests/`.

This is modeled after similar design in https://github.com/openshift/cluster-kube-apiserver-operator

Should unblock https://github.com/openshift/cluster-monitoring-operator/pull/391

/cc @sttts @s-urbaniak


paulfantom:
Removed `jobLabel: null` which is possibly causing a ServiceMonitor object to b invalid.

PR #154, opened 2019-07-17
Crd schema gen

ravisantoshgudimetla:
Built on top of https://github.com/openshift/cluster-kube-scheduler-operator/pull/152, this should pass tests

/cc @damemi @sttts


ravisantoshgudimetla:
@damemi - Unit tests are passing. The issue is with relying on featuregates which are have changed.

Force LGTM'ing depending on

https://github.com/openshift/cluster-kube-scheduler-operator/pull/152#issuecomment-510836532

PR #155, opened 2019-07-23
Fix to have hardcoded configmap name in the scheduler config

ravisantoshgudimetla:
Harcoded configmap name should be used instead of configmap name we get from scheduler CR.

/cc @damemi

### DIFF
--- a/pkg/operator/configobservation/scheduler/observe_scheduler.go
+++ b/pkg/operator/configobservation/scheduler/observe_scheduler.go
@@ -89,7 +89,7 @@ func ObserveSchedulerConfig(genericListers configobserver.Listers, recorder even
 		return prevObservedConfig, errs
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -89,7 +89,7 @@ func ObserveSchedulerConfig(genericListers configobserver.Listers, recorder even
 		return prevObservedConfig, errs
 	}
 
-	if err := unstructured.SetNestedField(observedConfig, configMapName, policyConfigMapNamePath...); err != nil {
+	if err := unstructured.SetNestedField(observedConfig, "policy-configmap", policyConfigMapNamePath...); err != nil {
```
ravisantoshgudimetla: Have it hardcoded here :)

PR #158, opened 2019-07-25
[release-4.1] Bug 1729918: Switch to configobserver pattern

damemi:
This backports changes in https://github.com/openshift/cluster-kube-scheduler-operator/pull/136/ (and subsequently https://github.com/openshift/cluster-kube-scheduler-operator/pull/155) to switch to the configobserver pattern, which happens to address config sync issues related to https://bugzilla.redhat.com/show_bug.cgi?id=1729918


jwforres:
1) I see no 4.2 BZ where QE validated this issue doesn't exist on 4.2 now
2) From your description it sounded speculative that this set of commits coming from multiple 4.2 PRs are what fixed the issue, can you confirm you manually repro'ed the bug on 4.1 without these commits, and then manually tested you could no longer repro the bug on a 4.1 cluster that has these commits?
3) Given the size of this change I would like to see approvals from your team and group lead on the necessity / risk / scope associated with this PR.  cc @sttts @mfojtik

damemi:
@jwforres thanks for looking at this. This bug was originally reported for 4.1.z so there wasn't a BZ for 4.2, though I did confirm in talks with @ravisantoshgudimetla that the issue was still present in 4.2. I've cloned the 4.1.z bug to target 4.2 and set in ON_QA so that we can confirm from qe that this is fixed.

I also verified in a 4.1 cluster that the bug existed, and that applying the patch from this PR fixed it. While we never identified the root cause of this issue, we did determine that it was related to something with configobservation and resource syncing and we were hopeful that the switch to this pattern would make the problem obsolete, which it did. It only happened to be 2 PRs due to a minor typo in the first that caused a regression.

jwforres:
/bugzilla refresh
since the bugs in question have changed

jwforres:
@damemi looks like your depends on between the two bugs is backwards, 4.1 should depend on 4.2

damemi:
@jwforres thanks, updated

eparis:
This is blocked because the 4.2 bug is still not complete and signed off by QE. @damemi you likely need to work on fixing this in 4.2 and getting QE signoff

damemi:
I have commented this week on the 4.2 bug (https://bugzilla.redhat.com/show_bug.cgi?id=1734486) with much more thorough details to confirm with QE that this *is* fixed and should be verified. Still waiting on them to actually verify but if you see my comments [starting here](https://bugzilla.redhat.com/show_bug.cgi?id=1734486#c10) I have explained that the previous failure to verify is actually due to misconfiguration in testing

PR #164, opened 2019-08-08
Bug 1736795: Add check for removal of policy configmap

damemi:
**TL;DR**: I did a lot of digging in this to find that it's actually dependent on an already-reported bug (https://bugzilla.redhat.com/show_bug.cgi?id=1738432), but we still need to make a minor adjustment to how we handle empty policy names in the config spec, which is the change here

-------------------------------------------

According to BZ https://bugzilla.redhat.com/show_bug.cgi?id=1736795, when the policy configmap name is set to `""` in the scheduler operator config (`oc get scheduler cluster`), this change doesn't seem to get carried over to the kubescheduler config (`oc get kubescheduler cluster`).

I think this is because we don't actually try to make any updates if the `len(Spec.Policy.Name)==0`, however the resourcesynccontroller *does* see this change, and subsequently deletes the synced policy configmap from the `openshift-kube-scheduler` namespace (this is what results in the error from the scheduler pods as described in the bug, `couldn't get policy config map openshift-kube-scheduler/policy-configmap: configmaps "policy-configmap" not found`). It would only fail on finding that configmap if it was looking for it, which it is because the kubescheduler cluster config doesn't update with an empty policy name.

However, trying to just clear the `observedConfig` and passing that in seems to silently fail at this update call: https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/vendor/github.com/openshift/library-go/pkg/operator/v1helpers/helpers.go#L131

I can tell that this is the spot that it fails because the `newSpec` looks like what I want:
```
&{ManagementState:Managed LogLevel: OperatorLogLevel: UnsupportedConfigOverrides:{Raw:[] Object:<nil>} ObservedConfig:{Raw:[] Object:<nil>}}
```
but the returned operator spec after that call still has a Raw value for observedconfig:
```
&{ManagementState:Managed LogLevel: OperatorLogLevel: UnsupportedConfigOverrides:{Raw:[] Object:<nil>} ObservedConfig:{Raw:[123 34 97 ... 125 125 125 125] Object:<nil>}}
```
(This also causes the operator to go into an infinite loop attempting to sync the new config)

I'm chasing this approach because I can confirm that as mentioned in the bug, manually deleting the `observedConfig` section from `oc edit kubescheduler cluster` when setting the `oc edit scheduler cluster` value for `Spec.Policy.Name` back to `""` does work, as expected.

So, I don't know if we can actually squash fields like this when trying to update the spec. I'm going to look more into it but it's here as a WIP and personal notepad

**Update:** the best I can find on this is that the above-mentioned UpdateOperatorSpec function calls the k8s dynamicClient `Update` function here: https://github.com/openshift/library-go/blob/a5507e7eb29c0ec68b73bd355cf5d248586fd776/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L81

In the k8s dynamicClient, this uses http `Put` here: https://github.com/kubernetes/client-go/blob/master/dynamic/simple.go#L147. I think if we want to clear the field entirely, this should be a json merge patch, but in my hacking changing to `Patch` still causes the same error. I have tried patching with both the `algorithmSource` field removed (via `RemoveNestedField`) and by setting `algorithmSource: nil`. 

It also seems that `mergo`, which is used in the dynamic client's call to [`getOperatorSpecFromUnstructured`](https://github.com/openshift/library-go/blob/a5507e7eb29c0ec68b73bd355cf5d248586fd776/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L134)` also does not overwrite (clear) existing fields when an empty field is passed to it. I suspect that *this* is the root cause of the issue, as it means that even if we set the rest call to use HTTP Patch, the object we are calculating to send does not consider the ability to clear fields


damemi:
/hold
until generic client is fixed, this will then require a bump of library-go

damemi:
/hold cancel
tested locally, including the client bump seems to fix this

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -82,6 +82,10 @@ func ObserveSchedulerConfig(genericListers configobserver.Listers, recorder even
 		sourceTargetLocation,
 	)
 	if len(configMapName) == 0 {
+		if len(currentPolicyConfigMapName) > 0 {
+			recorder.Eventf("ObservedConfigMapNameChanged", "scheduler configmap removed")
```
ravisantoshgudimetla: Can you update the existing unit test or add e2e for this scenario?

damemi: updated

pkg/operator/configobservation/scheduler/observe_scheduler_test.go
```diff
@@ -23,14 +23,22 @@ func TestObserveSchedulerConfig(t *testing.T) {
 	tests := []struct {
 		description   string
 		configMapName string
+		found         bool
```
ravisantoshgudimetla: nit: perhaps changing the name to appropriate variable name like `updateNeeded`?

damemi: updated

PR #165, opened 2019-08-15
Update OWNERS file to add new reviewers/approvers

damemi:
Adds me, Maciej, and Jan to owners as reviewers/approvers

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -3,8 +3,13 @@ reviewers:
   - sjenning
   - sttts
   - damemi
+  - soltysh
+  - ingvagabund
 approvers:
   - mfojtik
   - ravisantoshgudimetla
   - sjenning
   - sttts
+  - damemi
+  - soltysh
+  - ingvagabund


damemi:
Do we need to remove anyone? Owners list is getting big and I don't know if everyone on it is actively working on this repo

sttts:
/lgtm
/approve
/hold

Unhold any time when the question about who to delete is clarified.

@sjenning @ravisantoshgudimetla  do you want to stay in the list?

ravisantoshgudimetla:
I am fine taking my name off but my original intention was to be on for few more weeks.

sttts:
> I am fine taking my name off but my original intention was to be on for few more weeks.

No hurry, we leave you there then :)

soltysh:
It's not that big, let's keep Ravi around for a few more weeks as mentioned before. When removing Ravi we can also remove Seth, since he's there as a team lead where scheduler used to be. Rest should be there as is.
/hold cancel

PR #167, opened 2019-08-29
Bug 1745671: use cached clients in resource syncer

soltysh:
/assign @mfojtik @damemi


damemi:
/lgtm
thanks for finding this

PR #169, opened 2019-09-06
Fix makefile to allow make images to work

soltysh:
/assign @damemi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-kube-scheduler-operator
 COPY . .
-RUN go build ./cmd/cluster-kube-scheduler-operator
+RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 RUN mkdir -p /usr/share/bootkube/manifests
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-kube-scheduler-operator
 COPY . .
-RUN go build ./cmd/cluster-kube-scheduler-operator
+RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.0:base
 RUN mkdir -p /usr/share/bootkube/manifests
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,15 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 # Exclude e2e tests from unit testing
 GO_TEST_PACKAGES :=./pkg/... ./cmd/...
 
+IMAGE_REGISTRY :=registry.svc.ci.openshift.org
+
 # This will call a macro called "build-image" which will generate image specific targets based on the parameters:
 # $0 - macro name
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
-# It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(GO_PACKAGE),./Dockerfile,.)
+# $1 - target name
+# $2 - image ref
+# $3 - Dockerfile path
+# $4 - context directory for image build
+$(call build-image,ocp-cluster-kube-scheduler-operator,$(IMAGE_REGISTRY)/ocp/4.2:cluster-kube-scheduler-operator, ./Dockerfile.rhel7,.)
 
 # This will call a macro called "add-bindata" which will generate bindata specific targets based on the parameters:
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -9,16 +9,15 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
...
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
-# It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(GO_PACKAGE),./Dockerfile,.)
+# $1 - target name
+# $2 - image ref
+# $3 - Dockerfile path
+# $4 - context directory for image build
+$(call build-image,ocp-cluster-kube-scheduler-operator,$(IMAGE_REGISTRY)/ocp/4.2:cluster-kube-scheduler-operator, ./Dockerfile.rhel7,.)
```
ingvagabund: From https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/vendor/github.com/openshift/library-go/alpha-build-machinery/make/targets/openshift/images.mk#L28:

```sh
define build-image
$(eval $(call build-image-internal,$(1),$(2),$(3),$(4)))
endef
```

```sh
# $1 - target name
# $2 - image ref
# $3 - Dockerfile path
# $4 - context
define build-image-internal
```

Which, once rendered, will define Makefile's `image-ocp-cluster-kube-scheduler-operator` target:
```sh
$ make help | grep image
image-ocp-cluster-kube-scheduler-operator
images
```

Once `make images` is run, one will get:
```sh
$ imagebuilder --allow-pull -t registry.svc.ci.openshift.org/ocp/4.2:cluster-kube-scheduler-operator -f ./Dockerfile.rhel7 .
```

Looks good up to the point of running `imagebuilder` command.

ingvagabund: Nit: `$(IMAGE_REGISTRY)/ocp/4.2:cluster-kube-scheduler-operator` could be put under another variable, e.g.:

```
IMAGE_REGISTRY :=registry.svc.ci.openshift.org
TARGET_OPERATOR_IMAGE :=$(IMAGE_REGISTRY)/ocp/4.2:cluster-kube-scheduler-operator
...
$(call build-image,ocp-cluster-kube-scheduler-operator,$(TARGET_OPERATOR_IMAGE), ./Dockerfile.rhel7,.)
```

to improve readability.

Makefile
```diff
@@ -9,16 +9,15 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 	targets/openshift/images.mk \
 )
 
-# Exclude e2e tests from unit testing
-GO_TEST_PACKAGES :=./pkg/... ./cmd/...
```
tnozicka: global vars are by convention at the start of the file grouped together

Makefile
tnozicka: I think it less complicated without that extra variable. variables are intended to be set from outside - you can't change component name, just the registry.

let's take https://github.com/openshift/oc/blob/master/Makefile#L45 as a reference and keep it mostly similar

Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-kube-scheduler-operator
 COPY . .
-RUN go build ./cmd/cluster-kube-scheduler-operator
+ENV GO_PACKAGE github.com/openshift/cluster-kube-scheduler-operator
```
tnozicka: drop it, autodetected

Dockerfile
```diff
@@ -1,7 +1,8 @@
...
+RUN go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit=$(git describe --long --tags --abbrev=7 --match 'v[0-9]*')" ./cmd/cluster-kube-scheduler-operator
```
tnozicka: make build

Dockerfile.rhel7
```diff
@@ -1,7 +1,8 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-kube-scheduler-operator
 COPY . .
-RUN go build ./cmd/cluster-kube-scheduler-operator
+ENV GO_PACKAGE github.com/openshift/cluster-kube-scheduler-operator
+RUN go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit=$(git describe --long --tags --abbrev=7 --match 'v[0-9]*')" ./cmd/cluster-kube-scheduler-operator
```
tnozicka: make build

Dockerfile.rhel7
```diff
@@ -1,7 +1,8 @@
 FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
```
tnozicka: i think this is technically .rhel8 as that's the base image is comes from, admiting it's preexisting

soltysh: :man_shrugging: I'm leaving this consistent with all the rest.

Dockerfile
tnozicka: this was just a hack and doesn't support ART builds

PR #171, opened 2019-09-06
Use new crd-gen makefile targets from library-go

damemi:
This depends on https://github.com/openshift/library-go/pull/521 to include crd-schema-gen


damemi:
/hold
@sttts this would be an example of how we can keep using the `update-codegen` target with the library-go build machinery included

damemi:
/hold cancel
This now pulls in the new crd-schema-gen make targets from library-go

PR #175, opened 2019-09-24
Switch to genericoperatorclient

soltysh:
/assign @ingvagabund

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -9,12 +9,10 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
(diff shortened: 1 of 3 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -82,12 +81,12 @@ func NewTargetConfigReconciler(
 }
 
 func (c TargetConfigReconciler) sync() error {
-	operatorConfig, err := c.operatorConfigClient.KubeSchedulers().Get("cluster", metav1.GetOptions{})
+	operatorSpec, _, _, err := c.operatorClient.GetStaticPodOperatorState()
```
ingvagabund: `operatorClient` client is initialized through:

```
operatorClient, _, _ := genericoperatorclient.NewStaticPodOperatorClient(ctx.KubeConfig, operatorv1.GroupVersion.WithResource("kubeschedulers"))
```

which under the hood creates an informer for `kubeschedulers` gvr. Also, `c.operatorClient.GetStaticPodOperatorState()` uses the informer and gets the same "kubescheduler" spec. So the change is equivalent in what resource gets returned. Though, using informers instead so the rate is decreased from `~1.42` to `~1.24`.

PR #176, opened 2019-09-24
Bug 1745671: Do not register event handlers twice for configmaps under openshift-kube-scheduler

ingvagabund:
Reducing the number of syncs of TargetConfigReconciler to half.

Rates reduced to:
* configmaps: ~2
* rolebindings: ~1
* secrets: ~0.5

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -68,7 +68,6 @@ func NewTargetConfigReconciler(
 	configInformer.Config().V1().FeatureGates().Informer().AddEventHandler(c.eventHandler())
 	// we react to some config changes
 	kubeInformersForNamespaces.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
-	kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 
 	// we only watch some namespaces
 	namespacedKubeInformers.Core().V1().Namespaces().Informer().AddEventHandler(c.namespaceEventHandler())


pkg/operator/target_config_reconciler.go
```diff
@@ -68,7 +68,6 @@ func NewTargetConfigReconciler(
 	configInformer.Config().V1().FeatureGates().Informer().AddEventHandler(c.eventHandler())
 	// we react to some config changes
 	kubeInformersForNamespaces.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
-	kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
```
soltysh: If my understanding is correct, you could even remove the line above, since we register even handler for all config maps from `namespacedKubeInformers` and that contains both `TargetNamespace` and `GlobalUserSpecifiedConfigNamespace`, compare https://github.com/openshift/cluster-kube-scheduler-operator/blob/a42fcb7538e6fecad9278d650b04aced7c4e88be/pkg/operator/starter.go#L56-L63

ingvagabund: There are only two event handler registrations for configmaps:
```go
namespacedKubeInformers.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
kubeInformersForNamespaces.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
```

The former one is already namespaced so we can't use it to watch for configmap changes under `operatorclient.GlobalUserSpecifiedConfigNamespace`. So even the second event registration for config maps needs to stay.

ingvagabund:
```go
GlobalUserSpecifiedConfigNamespace    = "openshift-config"
TargetNamespace                       = "openshift-kube-scheduler"
```

PR #191, opened 2019-12-02
Bug 1771810: Add e2e to verify kso metrics accessible

sallyom:
example #2 for using library-go/test/metrics functions.  This PR adds an e2e to verify  cluster-kube-scheduler-operator metrics are accessible. 

see https://github.com/openshift/library-go/pull/602


ingvagabund:
>  This PR adds an e2e to verify cluster-kube-scheduler-operator metrics are accessible.

Is it required to access the metrics through kubectl? Or, is the goal of the PR to verify (using any method) the metrics are accessible? If so, would it be easier to access the metrics from the metrics endpoint? (I am just asking, I have never consumed the metrics other than directly from a component while running in the same binary).

sallyom:
> > This PR adds an e2e to verify cluster-kube-scheduler-operator metrics are accessible.
> 
> Is it required to access the metrics through kubectl? Or, is the goal of the PR to verify (using any method) the metrics are accessible? If so, would it be easier to access the metrics from the metrics endpoint? (I am just asking, I have never consumed the metrics other than directly from a component while running in the same binary).

The goal is to access metrics and also verify cluster CA -this will eventually be used to verify successful cert rotation- what I landed upon (after days of trial and error) is to access metrics via internal service url with curl cmd secured via cluster CA (generated by service-ca-operator/annotated configmap).  I also had a version using the external route, and go code instead of curl - but I could not utilize service CA - only way that worked was to set InsecureSkipVerify in TLSconfig - but exec-ing in a pod using internal url works perfectly. I'm open to using anything that works, though!!

ingvagabund:
> I'm open to using anything that works, though!!

I have zero experience in this area. Though, as long as it's used in tests, I have no objection. Are you aware of other components/operator that test the same? Or, are you pioneering it right now? :)

s-urbaniak:
@sallyom @ingvagabund hello from the monitoring team! :-) :wave: @paulfantom asked me to help out here, so don't hesitate also to ping me on #forum-monitoring in slack.

I do recommend that you access your metrics via the exposed Thanos routes (which internally forward to Prometheus). That way you will verify that your metrics are being successfully scraped and hence also verifies the RBAC setup towards your scraping targets. I would not advise to check metrics in your local metrics endpoint only.

Regarding querying metrics you could do the approach done here or create your own http client (as we did) after finding out the external route and token: https://github.com/openshift/cluster-monitoring-operator/blob/8e5623597b94d2f62c38bc427c23a820a01f002f/test/e2e/framework/route_client.go#L47-L76

sallyom:
> Regarding querying metrics you could do the approach done here or create your own http client (as we did) after finding out the external route and token: https://github.com/openshift/cluster-monitoring-operator/blob/8e5623597b94d2f62c38bc427c23a820a01f002f/test/e2e/framework/route_client.go#L47-L76

thanks @s-urbaniak yes, I was under impression that a TLSconfig secured via CAbundle was required, but as it turns out it is not - so, I am going w/ the external prom url /route client - i had that working a few days ago but only w/ InsecureSkipVerify - and I thought we needed to avoid that but alas after much discussion we do not - so I'm back to using the route (very similar to the route_client setup you pointed to).  That will live in library-go/test/library/metrics/query.go.

s-urbaniak:
@sallyom sounds good, yes we are also "InsecureSkipVerify'ing" the TLS connection via the exposed routes right now.

Makefile
```diff
@@ -34,6 +34,7 @@ $(call build-image,ocp-cluster-kube-scheduler-operator,$(IMAGE_REGISTRY)/ocp/4.2
 $(call add-bindata,v4.1.0,./bindata/v4.1.0/...,bindata,v410_00_assets,pkg/operator/v410_00_assets/bindata.go)
 
 e2e: GO_TEST_PACKAGES :=./test/e2e
+e2e: GO_TEST_FLAGS += -v
```
soltysh: I don't think we need that verbosity, if the test fails, we'll get it, otherwise it's not needed.

test/e2e/scheduler_test.go
```diff
@@ -254,6 +274,50 @@ func TestPolicyConfigMapUpdate(t *testing.T) {
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("prometheus returned unexpected results: %v", err)
+	}
+
+	// do something with result, eventually this will be tested b4 and after rotation to ensure values differ
+	fmt.Printf("result from prometheus query `scheduler_scheduling_duration_seconds_sum`: %v", response)
```
soltysh: Maybe change this to `t.Logf`

Makefile
sallyom: removed it

test/e2e/scheduler_test.go
sallyom: done

PR #193, opened 2019-12-11
update for library-go condition fixes

deads2k:


damemi:
@deads2k could you please add a link to what condition fixes this is pulling in (PR)?

deads2k:
minimally https://github.com/openshift/library-go/pull/632 https://github.com/openshift/library-go/pull/639 https://github.com/openshift/library-go/pull/638

PR #195, opened 2019-12-18
Bump: for library-go with go mod based verify-deps

tnozicka:
/cc @mfojtik @soltysh


vendor/github.com/openshift/library-go/alpha-build-machinery/make/targets/openshift/deps-gomod.mk
```diff
@@ -0,0 +1,41 @@
+self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
```
tnozicka: here it goes

go.mod
```diff
@@ -4,33 +4,27 @@ go 1.12
...
+	k8s.io/klog v1.0.0
 )
 
 replace (
 	github.com/jteeuwen/go-bindata => github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
 	github.com/openshift/api => github.com/openshift/api v3.9.1-0.20191209132752-992bc3a41fe6+incompatible
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191205152420-9faca5198b4f
-	github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20191211124107-e0f1590a316e
+	github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20191218095328-1c12909e5923
+	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.2
```
tnozicka: pinned the prometheus client back to avoid incompatible changes

soltysh: Yeah, I think @sallyom will get to it when she brings new k8s version.

PR #196, opened 2020-01-07
Bug 1788151: fix go.mod and bump library-go to pickup ipv6 change

mfojtik:


go.mod
```diff
@@ -19,19 +19,4 @@ require (
 	k8s.io/klog v1.0.0
 )
 
-replace (
-	github.com/jteeuwen/go-bindata => github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
```
ingvagabund: go.mod in library-go [1] is now pinning the revisions, correct?

[1] https://github.com/openshift/library-go/blob/master/go.mod

mfojtik: it is not, but this one does not seems to break nor need to be pinned?

ingvagabund: Dunno, though it can be always undone when needed. Just wondering what's better. Though, as long as the latest are compatible and acceptable, no nit here.

PR #204, opened 2020-02-06
bug 1798725: static pod operator cannot be composed

deads2k:
Update to fix the operator. This cannot be done in the static pod code itself because some consumers like etcd still rely on those directly set conditions and removing them would break it.


soltysh:
Rebased version in https://github.com/openshift/cluster-kube-scheduler-operator/pull/206

PR #205, opened 2020-02-06
Build machinery

mfojtik:


soltysh:
Already in the queue in https://github.com/openshift/cluster-kube-scheduler-operator/pull/203

PR #206, opened 2020-02-07
Bug 1798725: static pod operator cannot be composed

soltysh:
Replaces https://github.com/openshift/cluster-kube-scheduler-operator/pull/204


soltysh:
/hold
it looks like there's some problem with this.

PR #207, opened 2020-02-10
Bump: openshift/build-machinery-go for #7

tnozicka:
ART builder are still on 1.13.4

picks up https://github.com/openshift/build-machinery-go/pull/7

/cc @mfojtik @soltysh


tnozicka:
@soltysh will pick it into https://github.com/openshift/cluster-kube-scheduler-operator/pull/206
/close

PR #216, opened 2020-03-06
Bug 1810008: Add cert syncer

tnozicka:
kube-scheduler is required to be operational if we are starting after downtime or after cert regeneration and all the regular pods are down. There is no operator to create new installer pods so we need scheduler up first. That also means it has to be able to reload refreshed certs. Kubelet also needs sdn to be running for installer pods to work as they use service network and sdn needs scheduler to run.

/hold
this needs client cert reload in origin first https://github.com/openshift/origin/pull/24646
 
/cc @deads2k @soltysh


tnozicka:
/hold cancel
https://github.com/openshift/origin/pull/24646 merged

/retest

PR #214, opened 2020-03-05
Bug 1810181: Remove wildcard COPY form Dockerfile

tnozicka:
It doesn't work with docker and buildah, only image builder.

/cc @soltysh @bparees

PR #223, opened 2020-03-06
[release-4.1] Bug 1810186: Remove wildcard COPY form Dockerfile

tnozicka:
backport of https://github.com/openshift/cluster-kube-scheduler-operator/pull/214

/cc @soltysh @bparees


bparees:
i'm surprised to see the file deletion, was it dead/obsolete?

tnozicka:
@bparees that was never used or hooked into CI, and there was no OKD so there is no point in pretending we have non rhel dockerfiles. We have removed it later on but never backported it.

bparees:
thanks.

overriding cherry pick and BZ requirement.  lgtm+approval can come from your team

And thanks for jumping on all these.

PR #226, opened 2020-03-16
Rebase to kube 1.18

damemi:


damemi:
/hold
this is currently a POC based on the prebase branches that have merged in library-go, api, and client-go

pkg/operator/configmetrics/configmetrics.go
```diff
@@ -48,3 +48,7 @@ func (m *configMetrics) Collect(ch chan<- prometheus.Metric) {
 		ch <- g
 	}
 }
+
+func (m *configMetrics) FQName() string {
+	return "cluster_master_schedulable"
```
soltysh: Maybe worth making this a constant rather than repeating it in two places?

damemi:
Currently pushed with the changes from https://github.com/openshift/library-go/pull/748 in my fork of library-go, need that to get in to the prebase branch

damemi:
Reverted https://github.com/openshift/cluster-kube-scheduler-operator/pull/226/commits/630f2e5c218e9fc842d493b96fcc0918b93e4d96 to see if that's breaking tests

soltysh:
> Reverted [630f2e5](https://github.com/openshift/cluster-kube-scheduler-operator/commit/630f2e5c218e9fc842d493b96fcc0918b93e4d96) to see if that's breaking tests

That'd be surprising but let's see :man_shrugging:

damemi:
/retest
I think that switch to k8smetrics was breaking it. The cluster was failing with some weird logs around prometheus that pointed to the area where we register that metric.

damemi:
/hold cancel
I think using the old metrics is still fine, but would like to switch at some point so we can inherit those methods

PR #227, opened 2020-03-17
Bug 1812583: Normalize CPU requests on masters

soltysh:
The ks uses less than 1% of master CPU in a reasonable medium
sized workload. The ks can be bursty in certain scenarios.

Given a 1 core per master baseline (since CPU is compressible and
shared), assign the ks roughly 1.5% of that core on each master
(1/20 of the kube-apiserver usage).


soltysh:
4.4 bug is 1812709

/cherrypick release-4.4

Issue #231, opened 2020-03-26
TraceAll loglevel should be 10


damemi:
Many of the core scheduling decisions are logged at a level `V(10)`. Because of this, we should consider the [TraceAll](https://github.com/openshift/api/blob/master/operator/v1/types.go#L94) operand log level to be 10 (instead of the recommended 8). Currently you need to manually edit the scheduler pod spec to get this value

PR #232, opened 2020-03-26
Bump TraceAll log level to v=10

damemi:
Fixes https://github.com/openshift/cluster-kube-scheduler-operator/issues/231

Many critical spots in the scheduler code log at level `V(10)`, which is higher than our current `TraceAll` standard of 8. That means that in order to get certain helpful messages, developers currently need to manually hack the scheduler deployment to pass a `--v=10` flag which is not ideal when they could just use the loglevel setting in the operator spec.

Examples of these logs are:
* Node scores when prioritizing: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/core/generic_scheduler.go#L650-L654
* Several plugins' debugging information:
    * https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/volumezone/volume_zone.go#L167
    * https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/resource_allocation.go#L70-L84
    * https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/interpodaffinity/filtering.go#L410-L414

Though there are fewer usages of `V(10)` than in the past, it is still the usual log level for important debugging information within the scheduler

### DIFF
--- a/pkg/operator/target_config_reconciler_v410_00.go
+++ b/pkg/operator/target_config_reconciler_v410_00.go
@@ -141,7 +141,9 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
 	case operatorv1.Trace:
 		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", 6))
 	case operatorv1.TraceAll:
-		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", 8))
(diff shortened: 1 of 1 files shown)


damemi:
cc @deads2k

deads2k:
/hold

@damemi you'll want to add a comment about why 10 is used.  Which messages are missing without it.

damemi:
@deads2k @soltysh updated the main description with specific examples and added a comment referencing this PR for more info, ptal again
/cherry-pick release-4.4

PR #241, opened 2020-04-24
Bug 1834852: Add make target to run operator with telepresence 

marun:
Once telepresence has been installed (see the run-telepresence.sh script for instructions), `make telepresence` will replace the operator's pod with a local process that has access to the same environment as the pod. This can speed up development since iteration doesn't rely on image build/push/pull cycle. See https://telepresence.io for more details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	targets/openshift/bindata.mk \
 	targets/openshift/images.mk \
 	targets/openshift/deps.mk \
+	targets/openshift/operator/telepresence.mk \
 )
 
 # Exclude e2e tests from unit testing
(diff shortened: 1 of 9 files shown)


marun:
@damemi @ingvagabund Is there something I'm missing in this PR that would be breaking `e2e-aws-operator`, or is it just flaking?

Makefile
```diff
@@ -40,3 +41,8 @@ e2e: test-unit
 clean:
 	$(RM) ./cluster-kube-scheduler-operator
 .PHONY: clean
+
+# Configure the 'telepresence' target
+# See vendor/github.com/openshift/build-machinery-go/scripts/run-telepresence.sh for usage and configuration details
+export TP_DEPLOYMENT_YAML ?=./manifests/0000_25_kube-scheduler-operator_06_deployment.yaml
+export TP_CMD_PATH ?=./cmd/cluster-kube-scheduler-operator
```
ingvagabund, line 48: Since you don't use `$(call telepresence ...)`, will both variables have any effect?

ingvagabund:
@marun I don't see how your changes might be related to `e2e-aws-operator `

/retest

ingvagabund:
Changes are scoped to the Makefile only. No customer impact.

/lgtm
/approve

Makefile
marun: Thank you for your review! The make target added by line 10 will be configured by these values.  I've tested it to work on this and other PRs adding the telepresence target.  Feel free to give it a try! :)

Issue #197, opened 2020-01-13
Support kube scheduler ComponentConfig


damemi:
I know right now we support policy config through the operator, but don't know what our status is on componentconfig. So making this issue to track that and look into it

damemi:
There was some discussion in our channel about whether we'll ever actually need to support CC. I asked in sig-scheduling and it seems that yes, someday policy config will be deprecated but CC is still a long way from GA. So, this is something we should keep in mind for our operator but more of a long term goal

damemi:
Upstream tracking issue to deprecate policy in favor of componentconfig: https://github.com/kubernetes/kubernetes/issues/87526

damemi:
As of 1.19, the `v1alpha1` scheduler config library is removed, which is what our operator depends on. We'll need to update our scheduler to use ComponentConfig by then

Currently, we observe the scheduler config by giving the scheduler a configmap that has the policy in it (see https://github.com/openshift/cluster-kube-scheduler-operator/blob/82b5c997d78cbc2cdf8a3b661f82f59b43cc04fa/pkg/operator/configobservation/scheduler/observe_scheduler.go#L17)

In `v1alpha2`, the `AlgorithmSource` field is deprecated, meaning we can no longer simply pass a configmap name that holds a `Policy` file. Instead, we'll need to pass that observed config directly as a [`SchedulerProfile`](https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/kube-scheduler/config/v1alpha2/types.go#L94). Note that this means documentation should be updated to reflect that the scheduler will now be configured by Plugin (filter/score/etc) componentconfig and NOT policy (predicates/priorities)

For reference, currently when you try to run a >1.19 kube-scheduler image with our operator you get the following error right away:
```
no kind "KubeSchedulerConfiguration" is registered for version "kubescheduler.config.k8s.io/v1alpha1" in scheme "k8s.io/kubernetes/pkg/scheduler/apis/config/scheme/scheme.go:30"
```

Updating our KubeSchedulerConfiguration to v1alpha2 (https://github.com/openshift/cluster-kube-scheduler-operator/pull/245) fixes it, but trying to pass a custom config results in this error:
```
[...]

PR #245, opened 2020-04-30
[wip] Switch to v1alpha2 componentconfig

damemi:
See https://github.com/openshift/cluster-kube-scheduler-operator/issues/197

As of 1.19 the v1alpha1 scheduler config will be removed. We need to update our operator to use the new componentconfig (and docs to reflect the policy/plugin config change)


damemi:
/hold
this shouldn't merge until we're on a 1.19 scheduler

PR #246, opened 2020-05-01
Bug 1830406: Have targetconfig controller use cm/default-ingress-cert

danehans:
[This commit](https://github.com/openshift/cluster-kube-controller-manager-operator/commit/55e685148796dca181afe4330cecc60ea598d954) updated the targetconfig controller for kube-controller-manager to use the "default-ingress-cert" configmap. The same should be done for cluster-kube-scheduler-operator.

__Note:__ "default-ingress-cert" ConfigMap contains the self-signed CA cert and the router's server cert.
 
/cc @Miciah


Miciah:
> /cherry-pick release-4.3

See https://github.com/openshift/cluster-kube-controller-manager-operator/pull/410#issuecomment-625493947; I don't think we need or want a 4.3 backport.

danehans:
Ack. I'll close the 4.3 backport. Thanks for the guidance.

PR #248, opened 2020-05-12
Bug 1813894: Stop using service ca from service account token

marun:
/cc @sttts @mfojtik @deads2k


marun:
/cc @damemi @ingvagabund 

Apologies, I should have cc'd you to start with. PTAL!

PR #250, opened 2020-05-26
Bug 1840173: Add boolean metric to report legacy policy config

damemi:
This adds a boolean `cluster_legacy_scheduler_policy` metric that reports whether the operator is trying to use a legacy `Policy` config to customize the scheduler configuration

This also changes the operator's metric's FQName to `cluster_kube_scheduler_operator` now that there are multiple metrics it's reporting.


damemi:
I think the prometheus failure in e2e-aws-operator is an unrelated flake, because it's failing in other PRs as well (see https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-scheduler-operator/249/pull-ci-openshift-cluster-kube-scheduler-operator-master-e2e-aws-operator/708)

PR #254, opened 2020-06-15
Wait for localhost kas to respond

soltysh:
Similar to https://github.com/openshift/cluster-kube-controller-manager-operator/pull/420

/assign @deads2k

PR #255, opened 2020-06-16
Switch to v1beta1 Scheduler config to bridge between Policy/Plugin APIs

damemi:
(Note: This is a 1.19 blocker, see https://github.com/openshift/cluster-kube-scheduler-operator/issues/197)

The scheduler has changed from the old style of config ("Policy", which was based on Predicates and Priorities) to a new config API ("ComponentConfig", which is based on Plugins such as Filter and Score). The Policy API is officially deprecated, and will eventually be removed entirely.

Our operator currently parses Policy configs (as created in [this docs example](https://docs.openshift.com/container-platform/4.1/nodes/scheduling/nodes-scheduler-default.html)) and adds a reference to the Policy config in the `KubeSchedulerConfiguration`, which combines Predicate/Priority config with overall scheduler config.

We will be able to follow a similar approach once we have fully migrated to Plugin/ComponentConfig, because ComponentConfig also holds settings for plugin "profiles" and general scheduler configuration. However, we cannot yet fully adopt the Plugins because some names and availability have changed in the transition from Predicates/Priorities, and this will require more investigation and docs updates. (There is currently a built-in transition layer to convert Policy configs to their matching Plugins, which is what we have technically been using and will continue to take advantage of in this PR).

This PR uses the (deprecated) `--policy-configmap` and `--policy-configmap-namespace` flags to update the scheduler pod with this information when the operator spec is changed. The policy configmap will still be copied by the resource sync controller under a known name to the target namespace. However this means that we no longer need to (or are even able to) configure the `algorithmSource` fields in the config observer.

When we eventually switch to full ComponentConfig/Plugin setup: ComponentConfig accepts the entire Plugin set (the equivalent of the current Policy config). So prior to this PR, we only merged the configmap name into the Policy config. In ComponentConfig, we will merge the entire "policy-configmap" contents into the CC (because CC does not simply accept a reference to the configmap).

### DIFF
--- a/bindata/v4.1.0/config/defaultconfig.yaml
+++ b/bindata/v4.1.0/config/defaultconfig.yaml
@@ -1,2 +1,2 @@
-apiVersion: kubescheduler.config.k8s.io/v1alpha1
+apiVersion: kubescheduler.config.k8s.io/v1beta1
 kind: KubeSchedulerConfiguration
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -51,6 +51,7 @@ type TargetConfigController struct {
 	configMapLister       corev1listers.ConfigMapLister
 	featureGateLister     configlistersv1.FeatureGateLister
 	featureGateCacheSync  cache.InformerSynced
+	configInformer        configinformers.SharedInformerFactory
 	// queue only ever has one item, but it has nice error handling backoff/retry semantics
 	queue workqueue.RateLimitingInterface
 }
@@ -74,6 +75,7 @@ func NewTargetConfigController(
 		configMapLister:       kubeInformersForNamespaces.ConfigMapLister(),
 		operatorClient:        operatorClient,
 		eventRecorder:         eventRecorder,
+		configInformer:        configInformer,
 
 		featureGateLister:    configInformer.Config().V1().FeatureGates().Lister(),
 		featureGateCacheSync: configInformer.Config().V1().FeatureGates().Informer().HasSynced,
@@ -237,7 +239,7 @@ func createTargetConfigController_v311_00_to_latest(c TargetConfigController, re
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "serviceaccount/localhost-recovery-client", err))
 	}
-	_, _, err = managePod_v311_00_to_latest(c.ctx, c.kubeClient.CoreV1(), c.kubeClient.CoreV1(), recorder, operatorSpec, c.targetImagePullSpec, c.operatorImagePullSpec, c.featureGateLister)
+	_, _, err = managePod_v311_00_to_latest(c.ctx, c.kubeClient.CoreV1(), c.kubeClient.CoreV1(), recorder, operatorSpec, c.targetImagePullSpec, c.operatorImagePullSpec, c.featureGateLister, c.configInformer)
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/kube-scheduler-pod", err))
 	}
@@ -276,7 +278,7 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
 	return resourceapply.ApplyConfigMap(client, recorder, requiredConfigMap)
 }
 
-func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1client.ConfigMapsGetter, secretsGetter corev1client.SecretsGetter, recorder events.Recorder, operatorSpec *operatorv1.StaticPodOperatorSpec, imagePullSpec, operatorImagePullSpec string, featureGateLister configlistersv1.FeatureGateLister) (*corev1.ConfigMap, bool, error) {
+func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1client.ConfigMapsGetter, secretsGetter corev1client.SecretsGetter, recorder events.Recorder, operatorSpec *operatorv1.StaticPodOperatorSpec, imagePullSpec, operatorImagePullSpec string, featureGateLister configlistersv1.FeatureGateLister, configInformer configinformers.SharedInformerFactory) (*corev1.ConfigMap, bool, error) {
 	required := resourceread.ReadPodV1OrDie(v410_00_assets.MustAsset("v4.1.0/kube-scheduler/pod.yaml"))
 	images := map[string]string{
 		"${IMAGE}":          imagePullSpec,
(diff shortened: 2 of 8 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -305,6 +308,12 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
 	allFeatureGates := getFeatureGateString(sortedFeatureGates, featureGates)
 	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("--feature-gates=%v", allFeatureGates))
 
+	if len(os.Getenv("SCHEDULER_IMAGE")) != 0 {
```
soltysh, line 311: What is this for?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -327,6 +336,15 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
 		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, "--tls-private-key-file=/etc/kubernetes/static-pod-resources/secrets/serving-cert/tls.key")
 	}
 
+	config, err := configInformer.Config().V1().Schedulers().Lister().Get("cluster")
```
soltysh, line 339: This is changing, see https://github.com/openshift/cluster-kube-scheduler-operator/pull/254

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
damemi: Oh, I should remove this. I needed this because there is a problem with simply trying to set `${IMAGE}` for a custom operand image when you build a kube-scheduler that isn't hyperkube. Because we use `${IMAGE}` in the [initcontainer](https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/bindata/v4.1.0/kube-scheduler/pod.yaml#L15) as well as the operand container our current logic breaks if you supply it a kube-scheduler image that doesn't have `wait-for-host-port`. Or it's something like that

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
damemi: @soltysh is this change just moving from a slice of args to a concatenated string?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
soltysh: :stuck_out_tongue_winking_eye:

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
soltysh: Yup

damemi:
/hold
on https://github.com/openshift/cluster-kube-scheduler-operator/pull/254

damemi:
Pushed a commit to remove some test functionality that was based on the old `algorithmSource` policy config.

Because we will ultimately need to re-implement this same config observer approach, I don't want to completely remove everything that's using it. The tests cases will still essentially be valid, so it's worth keeping the bones of the tests and configobservers to guide re-implementation (following the todo's I've added)

To clarify, ComponentConfig accepts the entire Plugin set (the equivalent of the current Policy config). So prior to this PR, we only merged the configmap name into the Policy config. In ComponentConfig, we will merge the entire "policy-configmap" contents into the CC (because CC does not simply accept a reference to the configmap).

bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -1,2 +1,2 @@
-apiVersion: kubescheduler.config.k8s.io/v1alpha1
```
ingvagabund, line 1: Just a question. Why still `bindata/v4.1.0` and not `bindata/v4.6.0`?

damemi: No idea, I think it's just to denote that this is v4+. See other components like https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/bindata/v4.1.0

soltysh:
@damemi give that #254 is on hold, let's continue with this as is and I'll rebase mine

bindata/v4.1.0/config/defaultconfig.yaml
soltysh: Historical, theoretically this could have any name. We initially anticipated that there will a need for upgrade and changing the versions of resource definition based on version, but that was never implemented.

damemi:
I think this is failing because the API we're switching to is only present in 1.19+ k8s. Unfortunately the one we're currently using is also removed from 1.19 k8s.

So without this change, trying to use the 1.19 scheduler with our operator will crashloop. But trying to apply the change without a 1.19 scheduler also crashloops. Seems like a deadlock, I'll keep looking to find if we can ratchet the change in somehow but I don't think there is one.

damemi:
The current api, `v1alpha1`, was removed in https://github.com/kubernetes/kubernetes/pull/89298 in favor of `v1alpha2` which was introduced in https://github.com/kubernetes/kubernetes/pull/87628 and removed in https://github.com/kubernetes/kubernetes/pull/91420 for `v1beta1`.

`v1alpha2` has the same incompatibility issue as `v1beta1`, because `algorithmSource` was removed from it. So if we could carry support for the removed `v1alpha1` api, I think that would let us land a rebase that includes `v1beta1`, and we can then make the change while we have both APIs supported.

ingvagabund:
> v1alpha2 has the same incompatibility issue as v1beta1, because algorithmSource was removed from it. So if we could carry support for the removed v1alpha1 api, I think that would let us land a rebase that includes v1beta1, and we can then make the change while we have both APIs supported.

+1, versioned KubeSchedulerConfiguration is stored into internal/unversioned KubeSchedulerConfiguration anyway so the transition should be clear.

damemi:
This depends on https://github.com/openshift/origin/pull/25203, to make v1beta1 available

damemi:
/retest
/hold
for temp api bump

damemi:
Note: this will depend on the origin PR to pass CI, see https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-scheduler-operator/255/pull-ci-openshift-cluster-kube-scheduler-operator-master-e2e-aws-upgrade/1277956597649248256/artifacts/e2e-aws-upgrade/pods/openshift-kube-scheduler_openshift-kube-scheduler-ip-10-0-141-234.us-west-2.compute.internal_kube-scheduler.log

damemi:
This is failing with
```
Jul 06 14:38:18 ip-10-0-57-76 bootkube.sh[226626]: Rendering Kubernetes Scheduler core manifests...
Jul 06 14:38:19 ip-10-0-57-76 bootkube.sh[226626]: F0706 14:38:19.415875       1 render.go:42] failed to generate bootstrap config (phase 1): failed to merge configs: kubescheduler.config.k8s.io/v1alpha1/KubeSchedulerConfiguration does not equal kubescheduler.config.k8s.io/v1beta1/KubeSchedulerConfiguration
```

so it seems like there is still something in the bootstrap phase that's using `v1alpha1`

damemi:
> This is failing with
> 
> ```
> Jul 06 14:38:18 ip-10-0-57-76 bootkube.sh[226626]: Rendering Kubernetes Scheduler core manifests...
> Jul 06 14:38:19 ip-10-0-57-76 bootkube.sh[226626]: F0706 14:38:19.415875       1 render.go:42] failed to generate bootstrap config (phase 1): failed to merge configs: kubescheduler.config.k8s.io/v1alpha1/KubeSchedulerConfiguration does not equal kubescheduler.config.k8s.io/v1beta1/KubeSchedulerConfiguration
> ```
> 
> so it seems like there is still something in the bootstrap phase that's using `v1alpha1`

I think this was the bootstrap overrides. I've also opened a stripped version of this PR in https://github.com/openshift/cluster-kube-scheduler-operator/pull/261 that doesn't add new Profile support but is just the first commits here

damemi:
Removed some commits from this PR so this it is only the critical changes for 1.19

damemi:
/hold cancel
@soltysh this is ready for final review. It is only the critical aspects necessary to get kso working for the 1.19 bump.

The operator-e2e is flaking on prometheus (I have seen it run many times on other PRs and eventually unstick). Don't think it's related to this change

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -327,6 +329,15 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
...
+	if err != nil {
+		return nil, false, err
+	}
+	if len(config.Spec.Policy.Name) > 0 {
+		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, "--policy-configmap=policy-configmap")
```
ingvagabund, line 337: `required` is always a fresh pod manifest so there should be no argument conflicts

ingvagabund:
/lgtm

Without v1alpha1 -> v1beta1 change this is an equivalent change of saying "use this configmap when setting custom policy".
Given v1beta1 is finally merged in our k8s kube-scheduler fork, it's ok to bump the version as well while still keeping backward compatibility.

PR #262, opened 2020-07-09
Parse CustomNoUpgrade feature gates

bertinatto:
Currently if we add a feature gate to the `featuregates.config.openshift.io/cluster` object, the _controller-manager_ and the _apiserver_ correctly enable the feature, but the scheduler does not.

This patch adds the missing code that handles this functionality.


bertinatto:
> Does that look reasonable? The `if/elseif/else` isn't quite as clean as using straight returns, but I think this would help keep this code more readable. I won't block on this though so I'll approve now, and if you want to push that change then someone can lgtm

Hi Mike, I'm fine with that. PTAL

PR #264, opened 2020-07-22
[hold] Test bump to upstream rc1 prebase branches

damemi:


damemi:
oh didn't know we were merging this one, weird that "[hold]" in the title didn't add the hold label...

deads2k:
only [wip] works

nice job guys....

damemi:
> only [wip] works
> 
> nice job guys....

Is that intentional/recent? I've definitely held PRs with [hold] before, it's a different label than wip too

soltysh:
> Is that intentional/recent? I've definitely held PRs with [hold] before, it's a different label than wip too

hold only works with `/` which you explicitly invoke, wip worked always

PR #265, opened 2020-07-22
Bump to k8s 1.19-rc2 

damemi:
Bump client-go, library-go, api to pick up their 1.19-rc2 merged to master and updates to klog/v2 to match them


damemi:
This prometheus test flakes so much, we should look at fixing it or removing it if it's no longer necessary

PR #268, opened 2020-07-31
Drop code that expects the config map in openshift-config namespace to have policy-configmap name

ingvagabund:
Copy pasting `openshift-config/CMName` to `openshift-kube-scheduler/policy-configmap` is already carried out in the [config observer code](https://github.com/openshift/cluster-kube-scheduler-operator/blob/2f782886a6ffb61e52c9c0aa5d8dc842eb85125f/pkg/operator/configobservation/scheduler/observe_scheduler.go#L47-L52). Also, resource sync controller only copy pastes the user specified config when it's `policy-configmap` named (`SyncConfigMap` has the first argument as a destination. the second one as a source).

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesynccontroller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesynccontroller.go
@@ -2,9 +2,7 @@ package resourcesynccontroller
 
 import (
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/klog/v2"
 
-	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	"github.com/openshift/cluster-kube-scheduler-operator/pkg/operator/operatorclient"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
@@ -14,7 +12,6 @@ import (
 func NewResourceSyncController(
 	operatorConfigClient v1helpers.OperatorClient,
 	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
-	configInformer configinformers.SharedInformerFactory,
 	kubeClient kubernetes.Interface,
 	eventRecorder events.Recorder) (*resourcesynccontroller.ResourceSyncController, error) {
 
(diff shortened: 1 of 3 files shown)


ingvagabund:
@damemi FYI

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -26,17 +23,6 @@ func NewResourceSyncController(
...
 	)
 
-	scheduler, err := configInformer.Config().V1().Schedulers().Lister().Get("cluster")
-	if err != nil {
-		klog.Infof("Error while listing scheduler %v", err)
-	}
-	if scheduler != nil && len(scheduler.Spec.Policy.Name) > 0 {
-		if err := resourceSyncController.SyncConfigMap(
-			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: scheduler.Spec.Policy.Name},
-			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalUserSpecifiedConfigNamespace, Name: "policy-configmap"}); err != nil {
```
damemi, line 36: I agree these lines are wrong (namespace/name are switched) but do we want to remove them?

Target config controller does basically the same copy, but only when the scheduler CR is updated. Do we have a test for when the user creates a custom policy configmap (in `openshift-config`), adds it to the scheduler CR, then later updates that custom policy configmap directly?

I am wondering if, without this, changes will only be synced between `openshift-config` and the target namespace when the scheduler CR is updated (so really only by changing the custom policy configmap name)

ingvagabund: > Target config controller does basically the same copy, but only when the scheduler CR is updated. 

Are you sure about this? I tested it locally and the configmap got synced after I edited openshift-config one even without the lines.

damemi: I actually wasn't sure, sorry, meant to pose that as more "did we check this?" based on talking about it offline the other day. If it works then cool! But do we have a test for that case?

ingvagabund: We do here https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/test/e2e/scheduler_test.go#L204

PR #278, opened 2020-09-04
[release-4.5] Bug 1874597: static-pod-installer: remove deadlock by recreating installers that disappear

soltysh:
…allers that disappear

/assign @sttts


damemi:
This prometheus flake is annoying, and I have no idea what causes it. We tried extending the timeout in https://github.com/openshift/cluster-kube-scheduler-operator/pull/271 but that clearly isn't enough:

```
 --- FAIL: TestMetricsAccessible (30.84s)
    scheduler_test.go:317: prometheus returned unexpected results: timed out waiting for the condition
FAIL
FAIL	github.com/openshift/cluster-kube-scheduler-operator/test/e2e	43.570s
FAIL
```

wangke19:
Launched one 4.5 cluster with this PR successfully, it worked fine.

markmc:
Risk assessment made difficult here because the critical 9 line fix we need ends up as a +749 −298 diff because of the necessary revendoring. Approving nonetheless, especially since we've merged this for ceo and ckao already.

PR #283, opened 2020-09-21
Bug 1907998: Collect /metrics/resources from scheduler

smarterclayton:
As part of KEP 1916, resource consumption from pods will be tracked by the scheduler and reported. This commit will allow that endpoint to be scraped once that code is available.  This enables those metrics to be scraped.  It adds roughly 1 series per pod on average (pods usually declare ~1 resource request on average).

From the e2e run, showing the sum of requested resources for the cluster

image

Subsequent releases will add resource capacity dashboards.

### DIFF
--- a/manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
+++ b/manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
@@ -1,4 +1,14 @@
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus-k8s-scheduler-resources
+rules:
+- nonResourceURLs:
+  - /metrics/resources
+  verbs:
+  - get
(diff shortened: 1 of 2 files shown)


smarterclayton:
This is ready for review @lilic / @s-urbaniak.  I want to add a recording rule that sums resource requests and returns for workloads on the cluster and will do that in a follow up PR.  Manually verified cardinality as described in the discussion in e2e run and verified the scrape returned the correct data.  I believe low risk for 4.7 but you're call.

lilic:
Nice! 🎉 I agree the risk is low here, an in the worst case we revert this change as we don't use these metrics in any rules yet.
/lgtm

/hold
@s-urbaniak you got the final call.

/retest

smarterclayton:
Second commit deals with - https://github.com/kubernetes/kubernetes/pull/94884 which in 1.20 removed the metric we test (just came up because i had the first PR to touch metrics)

PR #284, opened 2020-09-23
[release-4.4] Bug 1876484: static-pod-installer: remove deadlock by recreating installers that disappear

soltysh:
…allers that disappear

Picks openshift/library-go#881

/assign @sttts @sdodson


markmc:
/retitle [release-4.4] Bug 1876484: static-pod-installer: remove deadlock by recreating installers that disappear

PR #303, opened 2020-11-03
Revert "Set dnsPolicy ClusterFirstWithHostNet to match hostNetwork"

soltysh:
This reverts commit 3ef0d47ec29778ad15c42aef50f33025a44dd0b1 (#293) which is causing that our static pods have in-cluster dns server injected (see https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/kubelet/network/dns/dns.go#L348-L352), which is problematic during upgrades. I'm talking with the node team how and if this is going to be solved in kubelet. 

/assign @damemi


damemi (approved):
/approve
/lgtm

PR #310, opened 2020-12-08
Bug 1905327: specify cpu and memory resources for initContainers

soltysh:
/assign @damemi


bindata/v4.1.0/kube-scheduler/pod.yaml
```diff
@@ -23,6 +23,11 @@ spec:
         echo -n "."
         sleep 1
       done
+
```
damemi, line 26: are these newlines intentional?

soltysh: Yes, it's to make this more readable

PR #311, opened 2020-12-17
Bug 1908145: Change recovery-controller port to avoid conflicts

damemi:
This changes the recovery-controller port to `11443`, to avoid conflicting with the router


damemi (commented):
/cc @soltysh @tnozicka 
is this a reasonable port, and are there any other places this needs to be updated? I'm not very familiar with the recovery controller

damemi:
bump
@tnozicka @soltysh does this look good to you?

tnozicka (approved):
/lgtm
/approve

PR #324, opened 2021-01-26
bug 1920509: Use ss instead of lsof when waiting for ports

ingvagabund:
"ss" shows sockets in CLOSE-WAIT state, so it's more accurate.

Following on https://github.com/openshift/cluster-kube-scheduler-operator/pull/252

CC @soltysh


soltysh (approved):
/lgtm

PR #325, opened 2021-01-26
Wait for localhost kas to respond

ingvagabund:
Following on https://github.com/openshift/cluster-kube-scheduler-operator/pull/254

### DIFF
--- a/bindata/v4.1.0/kube-scheduler/pod.yaml
+++ b/bindata/v4.1.0/kube-scheduler/pod.yaml
@@ -24,6 +24,12 @@ spec:
         sleep 1
       done
 
+      echo -n "Waiting for kube-apiserver to respond."
+      until curl --output /dev/null --silent -k https://localhost:6443/readyz; do
+        echo -n "."
+        sleep 1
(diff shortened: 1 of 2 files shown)


damemi (commented):
/lgtm
looks like the difference between this and the old PR is that this does the check in the initContainers, right?

ingvagabund:
/hold
for discussion first

bindata/v4.1.0/kube-scheduler/pod.yaml
```diff
@@ -24,6 +24,12 @@ spec:
         sleep 1
       done
 
+      echo -n "Waiting for kube-apiserver to respond."
+      until curl --output /dev/null --silent -k https://localhost:6443/readyz; do
```
deads2k, line 28: This looks improper to me.  We switched to the internal LB for scheduler communication specifically to avoid this linkage.

ingvagabund:
/close
per https://github.com/openshift/cluster-kube-scheduler-operator/pull/325#discussion_r564548743

PR #330, opened 2021-02-11
Bug 1926579: Move warning about deprecated spec.policy field

damemi:
The operator must currently observe configmaps in the target namespace, however some (like leader-election) update frequently, triggering event handling in the targetConfigController and this log message to be printed every ~3s.

Moving it to only be logged on actual updates to the relevant policy configmap (and by extension the scheduler pod itself) will limit how often this is logged.

The alternative to this was changing the controller's event handlers to ignore updates from certain specific resources, but that leads to too many edge cases and the likelihood of regression if new targetNamespace resources are added.


damemi (commented):
/cc @soltysh @ingvagabund

PR #331, opened 2021-02-18
bug 1844989: Allow to set --tls-cipher-suites and --tls-min-version of KS instances through observed config

ingvagabund:
By allowing to set the ciphers, KS does not have to use the default ciphers some of which may contain security vulnerabilities.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
+
 	v1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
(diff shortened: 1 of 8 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -304,7 +307,7 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(lister corev1listers.ConfigM
 		}
 	}
 
-	requiredConfigMap, _, err := resourcemerge.MergeConfigMap(configMap, "config.yaml", nil, kubeSchedulerConfiguration, operatorSpec.ObservedConfig.Raw, operatorSpec.UnsupportedConfigOverrides.Raw)
+	requiredConfigMap, _, err := resourcemerge.MergeConfigMap(configMap, "config.yaml", nil, kubeSchedulerConfiguration, operatorSpec.UnsupportedConfigOverrides.Raw)
```
ingvagabund, line 310: @damemi I am removing `operatorSpec.ObservedConfig.Raw` from the merging op. It's quite dangerous to have it here as depending on the observers, invalid data can be written into `kubescheduler.config.k8s.io/v1beta1`. The same holds for `operatorSpec.UnsupportedConfigOverrides.Raw`. I guess it was used to pass location of the policy file into the config. Though, it's no longer a case. Also, I am not aware of any other observed bits suitable for `kubescheduler.config.k8s.io/v1beta1`.

damemi: Sounds good to me, yeah that was used to merge in the policy file name before and is no longer used.

damemi (commented):
/lgtm

PR #332, opened 2021-02-19
bug 1844989: Set --tls-cipher-suites, resp. --tls-min-version only when the fields under servingInfo observedConfig are found

ingvagabund:
In case `servingInfo.[cipherSuites|minTLSVersion]` do not exist under observedConfig bits

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -379,21 +379,21 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
 		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
 	}
 
-	cipherSuites, _, err := unstructured.NestedStringSlice(observedConfig, "servingInfo", "cipherSuites")
+	cipherSuites, cipherSuitesFound, err := unstructured.NestedStringSlice(observedConfig, "servingInfo", "cipherSuites")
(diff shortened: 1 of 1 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -379,21 +379,21 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
...
 	}
 
-	minTLSVersion, _, err := unstructured.NestedString(observedConfig, "servingInfo", "minTLSVersion")
+	minTLSVersion, minTLSVersionFound, err := unstructured.NestedString(observedConfig, "servingInfo", "minTLSVersion")
 	if err != nil {
 		return nil, false, fmt.Errorf("couldn't get the servingInfo.minTLSVersion config from observedConfig: %v", err)
 	}
 
-	if len(cipherSuites) > 0 {
+	if cipherSuitesFound && len(cipherSuites) > 0 {
```
damemi, line 392: is `len(cipherSuites) > 0` when it's not found?

ingvagabund: The code of `NestedStringSlice` does not suggests the assumption is always true.

damemi: Interesting, worst case I guess it doesn't hurt to check anyway
/lgtm

soltysh (approved):
/lgtm

PR #340, opened 2021-03-29
servicemonitor: Relabel kube_pod_resource_* pod/namespace labels

smarterclayton:
The tenancy proxy does not allow someone to fetch metrics for pods
unless they are namespace labelled. This is also a usability improvement
for end users.

### DIFF
--- a/manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
+++ b/manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
@@ -86,6 +86,7 @@ spec:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: scheduler.openshift-kube-scheduler.svc
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    honorLabels: true
     interval: 30s
     path: /metrics/resources
     port: https


manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
```diff
@@ -93,6 +93,13 @@ spec:
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: scheduler.openshift-kube-scheduler.svc
+    metric_relabel_configs:
```
simonpasquier, line 96: you could use [`honorLabels: true`](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#endpoint) instead.

smarterclayton: nice

smarterclayton:
Manually verified the labels are correct and the labels aren't lost.  PTAL

damemi (commented):
/approve
/retest

ingvagabund:
@simonpasquier is this good enough? I am not familiar with the field.

ingvagabund:
/lgtm
/hold

@simonpasquier please unhold if the change in the PR is reasonable

smarterclayton:
/hold cancel

yes it is (he suggested)

PR #341, opened 2021-03-29
 update library-go to fix tight loops

deads2k:
https://github.com/openshift/library-go/pull/1024 and https://github.com/openshift/library-go/pull/1036


soltysh (approved):
/lgtm
/approve

PR #343, opened 2021-04-08
Add management workload annotations

lack:
In support of the workload partitioning feature
(https://github.com/openshift/enhancements/pull/703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-scheduler-pod.yaml
@@ -8,6 +8,7 @@ metadata:
     openshift.io/component: "scheduler"
   annotations:
     openshift.io/run-level: "0"
+    workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
 spec:
   containers:
   - name: kube-scheduler
--- a/bindata/bootkube/manifests/00_openshift-kube-scheduler-ns.yaml
+++ b/bindata/bootkube/manifests/00_openshift-kube-scheduler-ns.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   annotations:
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
   name: openshift-kube-scheduler
   labels:
     openshift.io/run-level: "0"
--- a/bindata/v4.1.0/kube-scheduler/ns.yaml
+++ b/bindata/v4.1.0/kube-scheduler/ns.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   annotations:
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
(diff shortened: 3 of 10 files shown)


ingvagabund:
LGTM

@lack based on https://github.com/openshift/enhancements/pull/703 this looks straightforward. Anything else we should keep in mind besides annotation pods and namespaces?

lack:
> LGTM
> 
> @lack based on [openshift/enhancements#703](https://github.com/openshift/enhancements/pull/703) this looks straightforward. Anything else we should keep in mind besides annotating pods and namespaces?

I'd appreciate a quick sanity check to make sure I didn't miss any dynamically generated pods, deployments, or daemonsets (they're sometimes tricky to find to an outsider like me), but the good news is that all the magic happens in Kubelet, CRI-O, and a special admission hook - Basically by adding these annotations we're declaring all these pods as "part of OpenShift Management", and systems configured to do so will take all those management pods and pin them to specific CPU cores so they won't interfere with customer workloads in constrained environments.

ingvagabund:
Many our operators when updating a static pod on a master node create installer pods. Wondering if they need to be annotated as well.

ingvagabund:
Are the annotations meant only for long running pods?

lack:
> Many our operators when updating a static pod on a master node create installer pods. Wondering if they need to be annotated as well.

It's a lot more important for long-running pods that short-lived ones, but honestly if there's anything that counts as "OpenShift Management" (ie, not a customer workload) we really want to try to segregate those pods to the specific CPUSet to guarantee we don't steal from customer workloads.

Do you know where these installer pods are created, generally?

ingvagabund:
So the installer pod template is under https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml. Then extended in https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/installer_controller.go#L767. So, one would need to pass optional annotations to it since not every controller utilizing InstallerController would want to set those. There are other place where the wiring of annotations needs to be done.

@deads2k is it worth allowing operators to pass annotations down to the static pod installer?

lack:
> So the installer pod template is under https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml. Then extended in https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/installer_controller.go#L767. So, one would need to pass optional annotations to it since not every controller utilizing InstallerController would want to set those. There are other place where the wiring of annotations needs to be done.
> 
> @deads2k is it worth allowing operators to pass annotations down to the static pod installer?

There's ongoing discussion about these installer pods in a slack thread; out-of-scope for this change :)

soltysh (approved):
/lgtm
/approve

PR #364, opened 2021-07-29
Bug 1986829: metrics: use client cert auth for metrics scraping

stlaz:
The CMO is now capable of using client cert auth to scrape the metrics endpoints. Use that in our components to reduce API load and allow metrics collection when the API is down - given further authz conditions are met.


damemi (approved):
/lgtm
/approve

PR #365, opened 2021-08-16
Bug 1984608: Set kube-scheduler leader election defaults

damemi:
This sets default LeaderElection values to better handle API server disruption as recommended in https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#handling-kube-apiserver-disruption


damemi (commented):
/cc @soltysh

damemi:
/hold
Something weird about how I tried to break out the common config in the first commit broke this in all the clusters I tried, but copying it to each profile seems to have fixed it. Want to examine more before merging but it seems like the 2nd commit works at least

damemi:
/hold cancel
tests are green now, this is ready to tag

ravisantoshgudimetla (approved):
/lgtm

Thank you for working on this @damemi

PR #370, opened 2021-09-16
Bug 2003540: bump library-go 

aojea:
Signed-off-by: Antonio Ojea <aojea@redhat.com>


ingvagabund:
@aojea do you plan to carry the same change to the master branch as well? Or, is the change already in the master?

damemi (commented):
@aojea the bug needs to change its target version to 4.9.0:
> expected the bug to target the "4.9.0" release, but it targets "4.9.z" instead

aojea:
> @aojea the bug needs to change its target version to 4.9.0:
> 
> > expected the bug to target the "4.9.0" release, but it targets "4.9.z" instead

yeah, I'm confused with the process, since 4.9.0 seems to be open only for blocker+ bugs , anyway, will switch to 4.9.0 to see if is accepted too

PR #377, opened 2021-11-15
Fix gomod

soltysh:
Followup to https://github.com/openshift/cluster-kube-scheduler-operator/pull/375

/assign @damemi


damemi (commented):
/lgtm
/approve

PR #378, opened 2021-11-23
Bug 2002300: Disable balancedAllocation and add weight for HighNodeUtilization profile

damemi:
This will likely need to be backported

This profile disables `LeastAllocated` but it does not disable `BalancedAllocation`. It also is setting `MostAllocated` with a weight of `0` (see https://bugzilla.redhat.com/show_bug.cgi?id=2002300#c4)

### DIFF
--- a/bindata/assets/config/defaultconfig-postbootstrap-highnodeutilization.yaml
+++ b/bindata/assets/config/defaultconfig-postbootstrap-highnodeutilization.yaml
@@ -6,5 +6,7 @@ profiles:
       score:
         disabled:
         - name: "NodeResourcesLeastAllocated"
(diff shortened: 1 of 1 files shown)


damemi (commented):
/cc @soltysh @ingvagabund 
ptal, right now this profile doesn't do much. Adjusted the weight of the `MostAllocated` plugin to have a much more dramatic effect

ingvagabund:
As mentioned in the BZ the NodeResourcesMostAllocated's weight is 0 by default, there's a high chance noone has been using the profile extensively. Given NodeResourcesMostAllocated is the core plugin, we might even go with 10. Though, without a proper measuring, even 5 is an improvement.
/lgtm

PR #381, opened 2021-11-23
[release-4.7] Bug 2026111: Disable balancedAllocation and add weight for HighNodeUtilization profile

damemi:
Backport of https://github.com/openshift/cluster-kube-scheduler-operator/pull/378


soltysh:
@damemi give that 4.7 is in maintenance I'd drop this change.

PR #390, opened 2022-01-04
Bug 2033751: Return Error when trying to use Scheduler Policy

damemi:
The scheduler's Policy api was removed in 1.23, see https://github.com/kubernetes/kubernetes/pull/105828

As an alternative we could also keep just using the default (lownodeutilization) here, in order to not break the scheduler on upgrades. In that case, log a message that we're using the default profile instead. Open to ideas

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -173,10 +173,10 @@ func manageKubeSchedulerConfigMap_v311_00_to_latest(ctx context.Context, client 
 	if err != nil {
 		return nil, false, err
 	}
-	// TOOD(jchaloup): remove the condition once the policy API is removed from the code
-	// Until that, ignore profiles if the policy API is used.
+	// Scheduler Policy API has been removed in k8s 1.23
+	// Still keep the CRD field for compatibility, but it does nothing now
 	if len(config.Spec.Policy.Name) > 0 {
-		kubeSchedulerConfiguration = bindata.MustAsset("assets/config/defaultconfig-postbootstrap-lownodeutilization.yaml")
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm

damemi:
/hold
discussing if this error is the right approach

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -301,7 +297,7 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
 	configMap.Data["version"] = version.Get().String()
 	appliedConfigMap, changed, err := resourceapply.ApplyConfigMap(ctx, configMapsGetter, recorder, configMap)
 	if changed && len(config.Spec.Policy.Name) > 0 {
```
ingvagabund, line 299: With `manageKubeSchedulerConfigMap_v311_00_to_latest` returning error when `len(config.Spec.Policy.Name) > 0`, the kube-scheduler container will keep crash looping since the configmap with the profile will be missing (assuming the `scheduler/cluster` object was specified with `.spec.policy.name` set during the cluster provisioning). I wonder if you took this case into account and whether it would make more sense to move `len(config.Spec.Policy.Name) > 0` check alongside the `config, err := configSchedulerLister.Get("cluster")` line at the beginning of the `managePod_v311_00_to_latest` function and return an error as well so the pod does not get created and kept crash looping until the policy field is cleared?

The case covers the bootstrapping phase in which (when `.spec.policy.name` is not empty) the installation fails. So it is unlikely any admin will update the `scheduler/cluster` object rather than running the installation again. So, the net benefit for the normal installation is quite low. On the other hand when the hypershift topology is used (with one cluster hosting many control planes), postponing the creation of the kube-scheduler pod might safe the step of debugging why the pod is crash looping (depending on how the control plan is provisioned).

ingvagabund: My comment is more for the debugging purposes than the functional/conceptual ones. The operator will go degraded when the policy field is set.

PR #391, opened 2022-01-04
bug 2037665: Set upgradeable=false if using Policy field

damemi:
To prevent upgrades from breaking due to Policy being removed in 4.10 (see https://github.com/openshift/cluster-kube-scheduler-operator/pull/390)

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -135,7 +135,7 @@ func createTargetConfigController_v311_00_to_latest(ctx context.Context, syncCtx
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/scheduler-kubeconfig", err))
 	}
-	_, _, err = managePod_v311_00_to_latest(ctx, c.kubeClient.CoreV1(), c.kubeClient.CoreV1(), syncCtx.Recorder(), operatorSpec, c.targetImagePullSpec, c.operatorImagePullSpec, c.featureGateLister, c.configSchedulerLister)
+	_, _, deprecatedPolicy, err := managePod_v311_00_to_latest(ctx, c.kubeClient.CoreV1(), c.kubeClient.CoreV1(), syncCtx.Recorder(), operatorSpec, c.targetImagePullSpec, c.operatorImagePullSpec, c.featureGateLister, c.configSchedulerLister)
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/kube-scheduler-pod", err))
 	}
@@ -159,6 +159,18 @@ func createTargetConfigController_v311_00_to_latest(ctx context.Context, syncCtx
 	}
 	if _, _, err := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(condition)); err != nil {
 		return true, err
+	}
+
+	if deprecatedPolicy {
+		upgradeableCondition := operatorv1.OperatorCondition{
+			Type:   "PolicyUpgradeable",
+			Status: operatorv1.ConditionFalse,
+			Reason: "PolicyFieldSpecified",
+			Message: fmt.Sprintf("scheduler.policy field is set which was deprecated and is to be removed in the next release"),
+		}
+		if _, _, err := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(upgradeableCondition)); err != nil {
+			return true, err
+		}
 	}
 
 	return false, nil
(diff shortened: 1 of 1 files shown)


ingvagabund:
/retitle bug 2037665: Set upgradeable=false if using Policy field

ingvagabund:
Tests failing with:

```
pkg/operator/targetconfigcontroller/targetconfigcontroller.go:256:3: not enough arguments to return
	have (nil, bool, error)
	want (*"k8s.io/api/core/v1".ConfigMap, bool, bool, error)
pkg/operator/targetconfigcontroller/targetconfigcontroller.go:264:3: not enough arguments to return
	have (nil, bool, error)
	want (*"k8s.io/api/core/v1".ConfigMap, bool, bool, error)
pkg/operator/targetconfigcontroller/targetconfigcontroller.go:277:4: not enough arguments to return
	have (nil, bool, error)
	want (*"k8s.io/api/core/v1".ConfigMap, bool, bool, error)
pkg/operator/targetconfigcontroller/targetconfigcontroller.go:283:3: not enough arguments to return
	have (nil, bool, error)
	want (*"k8s.io/api/core/v1".ConfigMap, bool, bool, error)
pkg/operator/targetconfigcontroller/targetconfigcontroller.go:288:3: not enough arguments to return
	have (nil, bool, error)
	want (*"k8s.io/api/core/v1".ConfigMap, bool, bool, error)
```

damemi (commented):
/cc @ingvagabund @soltysh 
something like this for 4.9?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -300,10 +310,12 @@ func managePod_v311_00_to_latest(ctx context.Context, configMapsGetter corev1cli
 	configMap.Data["forceRedeploymentReason"] = operatorSpec.ForceRedeploymentReason
 	configMap.Data["version"] = version.Get().String()
 	appliedConfigMap, changed, err := resourceapply.ApplyConfigMap(ctx, configMapsGetter, recorder, configMap)
+	deprecatedPolicy := false
 	if changed && len(config.Spec.Policy.Name) > 0 {
+		deprecatedPolicy = true
```
ingvagabund, line 315: **Observation**: The scheduler operator registers the event handler for `config.openshift.io/v1/scheduler`. When `spec.Policy.Name` field in the `scheduler/cluster` object is updated, the `sync` method is triggered and `managePod_v311_00_to_latest` is invoked.
  - If the `managePod_v311_00_to_latest` function returns an error, the operator condition `TargetConfigControllerDegraded` is set to true.
  - If the `managePod_v311_00_to_latest` function does not return an error, the `changed` is set to true and the `deprecatedPolicy = len(config.Spec.Policy.Name) > 0`.

So when the `spec.Policy.Name` field is non-empty, either
- `condition/TargetConfigControllerDegraded=true`
- or `condition/OperatorStatusTypeUpgradeable=false`.

Assuming the cluster upgrade requires all operators to have
- `condition/TargetConfigControllerDegraded=false`
- and `condition/OperatorStatusTypeUpgradeable=true`

this change will stop the upgrade from happening when the policy API is in use (assuming the operator status will get updated on time).

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -161,6 +161,16 @@ func createTargetConfigController_v311_00_to_latest(ctx context.Context, syncCtx
 		return true, err
 	}
 
+	if deprecatedPolicy {
+		upgradeableCondition := operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeUpgradeable,
+			Status: operatorv1.ConditionFalse,
+		}
+		if _, _, err := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(upgradeableCondition)); err != nil {
```
ingvagabund, line 169: Would you mind updating both conditions in a single call?

ingvagabund (changes requested):

ingvagabund:
The approach looks good!!!

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
damemi: Are you saying that something with `spec.Policy.Name` would cause `managePod` to return an error? Or just observing the possible cases where `managePod` returns an error *before* we actually set `deprecatedPolicy = true`?

ingvagabund: I wanted to verify the operator will properly report a condition disallowing the cluster upgrade by checking possible cases of when the `spec.Policy.Name` is updated. I did not find any suspicious line of code which would claim otherwise.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -161,6 +161,16 @@ func createTargetConfigController_v311_00_to_latest(ctx context.Context, syncCtx
 		return true, err
 	}
 
+	if deprecatedPolicy {
+		upgradeableCondition := operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeUpgradeable,
+			Status: operatorv1.ConditionFalse,
```
soltysh, line 167: This should be similar to here: https://github.com/openshift/cluster-kube-apiserver-operator/blob/5435d65b520191ba74fb2cd8c84a4f5bdf604178/pkg/operator/certrotationtimeupgradeablecontroller/certrotationtime_upgradeable.go#L68-L71 so that we have clear information what caused it and how to prevent it. I'd imagine:
```golang
upgradeableCondition := operatorv1.OperatorCondition{
    Type: "PolicyUpgradeable",
    Status:  operatorv1.ConditionFalse,
	Reason:  "PolicyFieldSpecified",
    Message: fmt.Sprintf("scheduler.policy field is set to point to %q configmap which was deprecated and is to be removed in the next release", policy.Name),
}
```
or something like that :wink:

soltysh (commented):
One nit wrt condition, rest looks great!

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
damemi: Updated to your example

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -161,6 +161,18 @@ func createTargetConfigController_v311_00_to_latest(ctx context.Context, syncCtx
 		return true, err
 	}
 
+	if deprecatedPolicy {
+		upgradeableCondition := operatorv1.OperatorCondition{
+			Type:   "PolicyUpgradeable",
+			Status: operatorv1.ConditionFalse,
+			Reason: "PolicyFieldSpecified",
+			Message: fmt.Sprintf("scheduler.policy field is set which was deprecated and is to be removed in the next release"),
```
soltysh, line 169: Nit: `deprecated scheduler.policy field is set, and it is to be removed in the next release`

soltysh (approved):
one nit and the unit and verify failures are real too

ingvagabund:
Continuing the work in https://github.com/openshift/cluster-kube-scheduler-operator/pull/400.

Thank you @damemi for starting the effort!!!
/close

PR #392, opened 2022-01-05
bug 2035927: Replace NodeResourcesLeastAllocated/NodeResourcesMostAllocated plugins with NodeResourcesFit

ingvagabund:
Based on https://kubernetes.io/docs/reference/scheduling/config/#scheduler-configuration-migrations
and https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/2458-node-resource-score-strategy#summary
both NodeResourcesLeastAllocated and NodeResourcesMostAllocated plugins
were removed from v1beta2 and replaced with NodeResourcesFit.

### DIFF
--- a/bindata/assets/config/defaultconfig-postbootstrap-highnodeutilization.yaml
+++ b/bindata/assets/config/defaultconfig-postbootstrap-highnodeutilization.yaml
@@ -2,11 +2,15 @@ apiVersion: kubescheduler.config.k8s.io/v1beta2
 kind: KubeSchedulerConfiguration
 profiles:
   - schedulerName: default-scheduler
+    pluginConfig:
+      - args:
+          scoringStrategy:
+            type: MostAllocated
+        name: NodeResourcesFit
     plugins:
       score:
         disabled:
-        - name: "NodeResourcesLeastAllocated"
         - name: "NodeResourcesBalancedAllocation"
         enabled:
-        - name: "NodeResourcesMostAllocated"
+        - name: "NodeResourcesFit"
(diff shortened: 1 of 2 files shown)


ingvagabund:
/hold
until I verify the fix

bindata/assets/config/defaultconfig-postbootstrap-highnodeutilization.yaml
```diff
@@ -2,11 +2,16 @@ apiVersion: kubescheduler.config.k8s.io/v1beta2
...
+    pluginConfig:
+      - args:
+          scoringStrategy:
+            type: MostAllocated
+        name: NodeResourcesFit
     plugins:
       score:
         disabled:
-        - name: "NodeResourcesLeastAllocated"
+        - name: "NodeResourcesFit"
```
damemi, line 13: does this need to be disabled? I think there were changes made so that you don't need to explicitly disable a default plugin to override it (eg, change weight) anymore

ingvagabund: I am not sure. I was trying to figure that out. The `NodeResourcesFit` plugin can have only a single `pluginConfig` configuration. So what you are saying makes sense.

ingvagabund: Removed.

damemi (commented):
/approve

soltysh (approved):
/lgtm

PR #398, opened 2022-01-13
Bug 2040132: drop --port flag

soltysh:
/assign @ingvagabund


ingvagabund:
Manifests under https://github.com/openshift/cluster-kube-scheduler-operator/tree/master/pkg/operator/targetconfigcontroller/testdata still have the `--port` flag set.

soltysh:
@ingvagabund should be fixed now

PR #400, opened 2022-01-19
bug 2037665: Set upgradeable=false if using Policy field

ingvagabund:
To prevent upgrades from breaking due to Policy being removed in 4.10 (see #390)

Building on top of https://github.com/openshift/cluster-kube-scheduler-operator/pull/391


soltysh (approved):
/lgtm

soltysh:
Overriding the valid-bug label, since we need that to block 4.10 upgrades

PR #402, opened 2022-01-24
bug 2037665: Remove policy upgradeable when policy field cleared

ingvagabund:
Extending https://github.com/openshift/cluster-kube-scheduler-operator/pull/400 to remove the `PolicyUpgradeable` condition when the policy name field was previously set but cleared later on.


soltysh (approved):
/lgtm
/approve
/label backport-risk-assessed

PR #410, opened 2022-02-03
Do not wait for a port which is no longer used by the scheduler

ingvagabund:
The insecure port 10251 was removed from the kube-scheduler codebase.


soltysh (approved):
/lgtm

PR #425, opened 2022-05-05
bug 2064024: Update library-go to 80f9619c2

pacevedom:
Relates to https://github.com/openshift/library-go/pull/1347, where we increase missing static pods timeout in case we are in single node.


pacevedom:
@ravisantoshgudimetla @soltysh PTAL

ingvagabund:
@pacevedom do you have a BZ that can be attached to this PR? It's getting required since feature freeze is closed.

ingvagabund:
I see there are other three components pulling this change. Worth reusing the same BZ

ingvagabund:
/retitle bug 2064024: Update library-go to 80f9619c2

PR #430, opened 2022-05-26
Bug 2062459: Introduce sync unit test

ravisantoshgudimetla:
Unit test which checks if the bootstrap config and the controlplane config are in sync or not. This helps in reducing the problems like 

https://bugzilla.redhat.com/show_bug.cgi?id=2062459


soltysh (approved):
/lgtm

PR #434, opened 2022-08-08
Add client certificate and key to service monitor

xueqzhan:
Added client certificates based on https://github.com/deads2k/openshift-enhancements/blob/master/enhancements/monitoring/client-cert-scraping.md

### DIFF
--- a/manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
+++ b/manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
@@ -85,6 +85,8 @@ spec:
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: scheduler.openshift-kube-scheduler.svc
+      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+      keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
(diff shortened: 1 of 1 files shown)


manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
```diff
@@ -85,6 +85,8 @@ spec:
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: scheduler.openshift-kube-scheduler.svc
+      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
```
ingvagabund, line 88: Should `bearerTokenFile` be removed alongside?

bergerhoffer:
Just testing out something (with @ingvagabund's approval :slightly_smiling_face:)
/lgtm

manifests/0000_90_kube-scheduler-operator_04_servicemonitor-scheduler.yaml
xueqzhan: Good question. When everything works, bearerTokenFile should be removed for service monitor. But as it turned out, kube-rbac-proxy is not yet coded/configured properly to handle client certificate authentication yet. I am in the process of fixing it. Original PR can go in without harm. But removing bearerTokenFile will have to be done separately after the fix in kube-rbac-proxy.

ingvagabund:
Left over from https://bugzilla.redhat.com/show_bug.cgi?id=1986829

PR #456, opened 2023-01-05
Revert "Drop log flags removed in k8s 1.26"

derekhiggins:
This results in historical log files being wiped on restart

This reverts commit 2d66da34ffea52fe012677d9e62a65ad6f080007.


deads2k:
/lgtm

no conflicts with other merges, putting master back to right for logging. Fixes can come after.

PR #457, opened 2023-01-05
remove use of deprecated klog flags

sanchezl:
Previously deprecated klog flags will be removed completely in kube 1.26.
Logs of bootstrap pods were previously using the --log-file flag, will be available in the cri-o container logs instead.


sanchezl:
/hold
need to solve unbound log file size

PR #459, opened 2023-01-19
WRKLDS-649: Guard pod set readiness probe endpoint explicitly

ingvagabund:
Pulling https://github.com/openshift/library-go/pull/1457 and setting the readiness endpoint to /healthz explicitly. There's currently no /readyz endpoint.


ingvagabund:
/hold
until https://github.com/openshift/library-go/pull/1457 merges

Issue #467, opened 2023-03-14
Security: we should disable pprof in production environment


lance5890:
1 The debugging endpoint /debug/pprof is exposed over the unauthenticated 10251 port
2 This debugging endpoint can potentially leak sensitive information

<img width="1154" alt="image" src="https://user-images.githubusercontent.com/30207518/224956280-281389c8-fbd4-4d13-b42c-9deab82c7527.png">

lance5890:
@ingvagabund @knelasevero

PR #468, opened 2023-03-14
disable debug pporf with unauthenticated port

lance5890:
[issue 467](https://github.com/openshift/cluster-kube-scheduler-operator/issues/467)

disable debug pporf with unauthenticated port


lance5890:
@ingvagabund @knelasevero

ingvagabund:
Based on https://kubernetes.io/docs/reference/config-api/kube-scheduler-config.v1beta3/#DebuggingConfiguration:

Field | Description
-- | --
enableProfiling [Required] bool | enableProfiling enables profiling via web interface host:port/debug/pprof/
enableContentionProfiling [Required] bool | enableContentionProfiling enables lock contention profiling, if enableProfiling is true.

Checking https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/config/v1alpha1/types.go#L60-L66:
```go
// DebuggingConfiguration holds configuration for Debugging related features.
type DebuggingConfiguration struct {
	// enableProfiling enables profiling via web interface host:port/debug/pprof/
	EnableProfiling *bool `json:"enableProfiling,omitempty"`
	// enableContentionProfiling enables block profiling, if
	// enableProfiling is true.
	EnableContentionProfiling *bool `json:"enableContentionProfiling,omitempty"`
}
```

There's no mention of either of the fields being required. Though, the defaulting in the scheduler code (https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/v1beta2/defaults.go#L162-L170) makes it enabled by default:

```go
// Enable profiling by default in the scheduler
if obj.EnableProfiling == nil {
	obj.EnableProfiling = pointer.Bool(true)
}

// Enable contention profiling by default if profiling is enabled
if *obj.EnableProfiling && obj.EnableContentionProfiling == nil {
	obj.EnableContentionProfiling = pointer.Bool(true)
}
```

ingvagabund:
Nice catch! Thank you
/approve
/lgtm

ingvagabund:
@lance5890 it could be a transient issue. Or, as you say, e.g. a missing check/condition in the library-go. Would you mind sharing the operator logs to see why the node is not progressing?

lance5890:
> @lance5890 it could be a transient issue. Or, as you say, e.g. a missing check/condition in the library-go. Would you mind sharing the operator logs to see why the node is not progressing?

1. I have not resolved this problem yet , when i found the solution, i will propose a PR
2. I have also proposed some PR in library-go ,and cluster-policy-controller, but u guys have not payed more attention to this

lance5890:
@ingvagabund  The failed two CI test has nothing to do with this fix, could we skip this ?

ingvagabund:
@lance5890 we are not in hurry. Better wait until the CI starts going green. In general we avoid overriding tests unless really needed.

lance5890:
@ingvagabund could we cherry-pick this PR to 4.12 ?

ingvagabund:
We are currently in a process of preparing 4.13 release. Only critical fixes are accepted now.

The change is quite straightforward.

/cherrypick release-4.13
/cherrypick release-4.12

lance5890:
> @lance5890 agree the change is worth backporting to both 4.13 and 4.12. Given you authored the commit, would you please open the corresponding backports?

ok, i will propose a new pr to 4.12 and 4.13

PR #471, opened 2023-03-20
OCPBUGS-10568: migrate to using lease objects for leader election

tkashem:


soltysh:
/close
in favor of https://github.com/openshift/cluster-kube-scheduler-operator/pull/476

Issue #473, opened 2023-03-24
Restrict access to metrics to localhost


lance5890:
<img width="753" alt="image" src="https://user-images.githubusercontent.com/30207518/227460007-85d66d05-f2e9-4f3a-9cf8-415ce595ed40.png">

1. The kube-scheduler metrics are exposed via http, so we should restrict this to localhost
2. The ocp prometheus get kube-scheduler metrics by servicemonitor via https
3. The kube-scheduler health check is on 10259 port, and the 10251 is not used by the health check

lance5890:
@ingvagabund @knelasevero

PR #474, opened 2023-03-24
bind http metrics to localhost

lance5890:
[issue 473](https://github.com/openshift/cluster-kube-scheduler-operator/issues/473)

Restrict access to http metrics

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -11,3 +11,5 @@ leaderElection:
   resourceLock: "configmapsleases"
 enableProfiling: false
 enableContentionProfiling: false
+metricsBindAddress: "127.0.0.1:10251"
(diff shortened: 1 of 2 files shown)


lance5890:
@ingvagabund @knelasevero

lance5890:
> The kube-scheduler code should no longer be exposing 10251 port at all. The insecure serving bits were removed through [kubernetes/kubernetes#96345](https://github.com/kubernetes/kubernetes/pull/96345). The provided configuration should fail the validation: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/validation/validation.go#L81-L83.

yeah, u are right, I am using the ocp 4.9 release, which corresponds to kubernetes 1.22.5,  and this version has not merged your PR

thanks, maybe I should upgrade the ocp

PR #477, opened 2023-03-27
OCPBUGS-7785: migrate to using lease objects for leader election

soltysh:
Manually picked https://github.com/openshift/cluster-kube-scheduler-operator/pull/476
/assign @tkashem


soltysh:
I'm overriding the valid-bug, since this will be critical for 4.13->4.14 upgrade.

PR #483, opened 2023-05-03
use standard featuregate library from API

deads2k:
This brings ks-o up to the standard featuregate consumption library.


deads2k:
on the path for extgernal ccm

PR #484, opened 2023-05-10
Remove featureset flag and use only the manifest

deads2k:
The manifest now exists in every install and can be used by bootstrapping. The installer passes the arg now, so we can update the operator to only honor the manifest and eliminate interpretation.

peer to https://github.com/openshift/cluster-kube-apiserver-operator/pull/1491


deads2k:
labelling for the cccmo bug

PR #516, opened 2023-12-10
OCPBUGS-18697: bump(library-go): guard controller: delay in some cases to avoid quick reconciliation

ingvagabund:
Proof for https://github.com/openshift/library-go/pull/1624


ingvagabund:
Will get resolved through https://github.com/openshift/cluster-kube-scheduler-operator/pull/510

PR #523, opened 2023-12-22
WIP: Change YAML style of scheduler-kubeconfig from human to compact JSON

omertuc:
(This is currently a WIP PR suggestion)

# Background

The `scheduler-kubeconfig` configmap managed by this operator contains a
YAML formatted `kubeconfig` string. As part of a particular novel form
of OCP deployment (image based upgrade/installation), we have to
programmatically modify the `server:` entry in that kubeconfig. This is
easily done by parsing the YAML, editing the server entry and
serializing the YAML back into the configmap.

# Issue

Every YAML serializer has its own opinions of what YAML should look
like. When we perform the procedure described above, we get a kubeconfig
that's slightly different than what this operator outputs (different
indentation, different quoting decisions). As a result, it eventually
causes the kube-scheduler-operator to issue a new revision of the
scheduler. Rolling out the new revision takes a few minutes, which we
would prefer to avoid.

# Solution

In order to solve this we suggest the change in this commit which would
be to format this kubeconfig as unopinionated compact JSON as opposed to
human readable YAML

# Alternative solutions

We could have a textual "search-and-replace" for the `server:` entry in
our code, but we would prefer to do it right by properly parsing,
editing then serializing it.

### DIFF
--- a/bindata/assets/kube-scheduler/kubeconfig-cm.yaml
+++ b/bindata/assets/kube-scheduler/kubeconfig-cm.yaml
@@ -4,23 +4,4 @@ metadata:
   name: scheduler-kubeconfig
   namespace: openshift-kube-scheduler
 data:
-  kubeconfig: |
-    apiVersion: v1
-    clusters:
-      - cluster:
-          certificate-authority: /etc/kubernetes/static-pod-resources/configmaps/serviceaccount-ca/ca-bundle.crt
-          server: $LB_INT_URL
-        name: lb-int
-    contexts:
-      - context:
-          cluster: lb-int
-          user: kube-scheduler
-        name: kube-scheduler
-    current-context: kube-scheduler
-    kind: Config
-    preferences: {}
-    users:
-      - name: kube-scheduler
-        user:
-          client-certificate: /etc/kubernetes/static-pod-certs/secrets/kube-scheduler-client-cert-key/tls.crt
-          client-key: /etc/kubernetes/static-pod-certs/secrets/kube-scheduler-client-cert-key/tls.key
+  kubeconfig: '{"apiVersion":"v1","clusters":[{"cluster":{"certificate-authority":"/etc/kubernetes/static-pod-resources/configmaps/serviceaccount-ca/ca-bundle.crt","server":"$LB_INT_URL"},"name":"lb-int"}],"contexts":[{"context":{"cluster":"lb-int","user":"kube-scheduler"},"name":"kube-scheduler"}],"current-context":"kube-scheduler","kind":"Config","preferences":{},"users":[{"name":"kube-scheduler","user":{"client-certificate":"/etc/kubernetes/static-pod-certs/secrets/kube-scheduler-client-cert-key/tls.crt","client-key":"/etc/kubernetes/static-pod-certs/secrets/kube-scheduler-client-cert-key/tls.key"}}]}'
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
@@ -192,26 +192,7 @@ func Test_manageKubeSchedulerConfigMap_v311_00_to_latest(t *testing.T) {
 	}
 }
 
-var defaultKubeconfigData = `apiVersion: v1
-clusters:
-  - cluster:
-      certificate-authority: /etc/kubernetes/static-pod-resources/configmaps/serviceaccount-ca/ca-bundle.crt
-      server: https://127.0.0.1:443
-    name: lb-int
-contexts:
-  - context:
-      cluster: lb-int
-      user: kube-scheduler
-    name: kube-scheduler
-current-context: kube-scheduler
-kind: Config
-preferences: {}
-users:
-  - name: kube-scheduler
-    user:
-      client-certificate: /etc/kubernetes/static-pod-certs/secrets/kube-scheduler-client-cert-key/tls.crt
-      client-key: /etc/kubernetes/static-pod-certs/secrets/kube-scheduler-client-cert-key/tls.key
-`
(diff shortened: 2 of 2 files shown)


ingvagabund:
> The scheduler-kubeconfig configmap managed by this operator contains a
YAML formatted kubeconfig string. As part of a particular novel form
of OCP deployment (image based upgrade/installation),

Would you please elaborate more on this novel form? 

>  we have to programmatically modify the server: entry in that kubeconfig.

Only KS-operator should be allowed to modify the kubeconfig. Otherwise, any subsequent revision roll out will undo your changes. Is there a card tracking your work?

omertuc:
>Would you please elaborate more on this novel form?

The tl;dr is we install SNO OCP, take a snapshot of the filesystem, and reuse that snapshot to quickly spin up new clusters, after a bit of reconfiguration

For mor information we're at `#wg-image-based-upgrades` on Red Hat Slack. [Internal HLD doc](https://docs.google.com/document/d/1nxLv77YOpfoP_kILJ2rjoVWQzNepIiELrbc1QzDhfZ0/edit#heading=h.xjvs1bwazipz)

> Only KS-operator should be allowed to modify the kubeconfig

True, for now we're going around changing stuff we "shouldn't", in the long term we're thinking of ways to make this more formal and have the components themselves support the changes we need

>Otherwise, any subsequent revision roll out will undo your changes. 

That's fine, since we're making the changes we need across the entire cluster's resources, so everything will be consistent and subsequent revisions should be OK

> Is there a card tracking your work?

Not for this particular thing, but in general our worked is tracked under [this](https://issues.redhat.com/issues/?jql=labels%20%3D%20IBU) Jira label

omertuc:
This change shouldn't make a difference in terms of this operator's functionality, but it will make our life's easier and our solution simpler, so if we could merge it, it would be nice. But it's not a must

ingvagabund:
>  When we perform the procedure described above, we get a kubeconfig that's slightly different than what this operator outputs (different indentation, different quoting decisions). As a result, it eventually causes the kube-scheduler-operator to issue a new revision of the scheduler. 

I wonder about the "it eventually causes the kube-scheduler-operator to issue a new revision of the scheduler." part. Is there a proof the different formatting is the cause to spin a new revision? Can you provide the operator logs to see what is causing the new revision? You should see something similar to
```
new revision X triggered by "required secret/localhost-recovery-client-token has changed,required configmap/config has changed,required configmap/sa-token-signing-certs has changed"
```
in the logs.

omertuc:
>Is there a proof the different formatting is the cause to spin a new revision? 

I haven't seen it directly, @mresvanis did

>Are you updating both scheduler-kubeconfig and scheduler-kubeconfig-X CMs? 

Yes

> So if you edit both CMs with the same formatted yaml (no matter how differently formatted it is) the deep equal comparison will produce a match.

True, but I guess some controller ([this](https://github.com/openshift/cluster-kube-scheduler-operator/blob/58929681daab0cd128865639e9c17a2fdf6a8063/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L420)?) notices the different format and re-applies with the styling from the bindata directory

ingvagabund:
> True, but I guess some controller ([this](https://github.com/openshift/cluster-kube-scheduler-operator/blob/58929681daab0cd128865639e9c17a2fdf6a8063/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L420)?) notices the different format and re-applies with the styling from the bindata directory, and that creates a diff

That's the other part. In the KSO case (https://github.com/openshift/cluster-kube-scheduler-operator/blob/58929681daab0cd128865639e9c17a2fdf6a8063/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L413-L417) we just read `assets/kube-scheduler/kubeconfig-cm.yaml` as it is and perform string replacement. Given `$LB_INT_URL` is the only string that gets replaced and KSO is the only component allowed to create/update `openshift-kube-scheduler` CM I strongly recommend to follow the same steps. We might extend the code base to create a new package that you can import in your solution and directly invoke the same code that renders the configmap. Thus, to avoid any issues with formatting and the config representation.

omertuc:
> > True, but I guess some controller ([this](https://github.com/openshift/cluster-kube-scheduler-operator/blob/58929681daab0cd128865639e9c17a2fdf6a8063/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L420)?) notices the different format and re-applies with the styling from the bindata directory, and that creates a diff
> 
> That's the other part. In the KSO case (
> 
> https://github.com/openshift/cluster-kube-scheduler-operator/blob/58929681daab0cd128865639e9c17a2fdf6a8063/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L413-L417
> 
> ) we just read `assets/kube-scheduler/kubeconfig-cm.yaml` as it is and perform string replacement. Given `$LB_INT_URL` is the only string that gets replaced and KSO is the only component allowed to create/update `openshift-kube-scheduler` CM I strongly recommend to follow the same steps. We might extend the code base to create a new package that you can import in your solution and directly invoke the same code that renders the configmap. Thus, to avoid any issues with formatting and the config representation.

> Given `$LB_INT_URL` is the only string that gets replaced and KSO is the only component allowed to create/update `openshift-kube-scheduler` CM I strongly recommend to follow the same steps

But from my POV all I have is an already "rendered" template, the kubeconfig I operate on doesn't have an easily recognizable $LB_INT_URL string that I can simply replace, so properly parsing it to edit the server URL makes sense. I could do some form of regex search and replace to find the existing URL line, but of course I would like to avoid that.

>We might extend the code base to create a new package that you can import in your solution and directly invoke the same code that renders the configmap. Thus, to avoid any issues with formatting and the config representation.

The tool we use for editing all of the cluster resources is not written in Go so a Go package would not be very helpful in our case.

>I strongly discourage anyone from modifying the CM through a different solution. This will make debugging more complicated as we don't assume other component modifying KSO owned files behind our back.

Ideally as a long term solution we should find a way to stop scattering the cluster-name and cluster-base-domain and all the URLs that are derived from them all over the place in OCP, so they can be easily changed in few places (ideally one) without causing a flood of reconciliations

PR #539, opened 2024-04-03
WRKLDS-1168: tolerate node-role.kubernetes.io/control-plane:NoExecute

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.


bindata/assets/kube-scheduler/pod.yaml
```diff
@@ -120,6 +120,9 @@ spec:
   priorityClassName: system-node-critical
   tolerations:
   - operator: "Exists"
+  - key: "node-role.kubernetes.io/control-plane"
+    operator: "Exists"
+    effect: "NoExecute"
```
ingvagabund, line 125: `- operator: "Exists"` tolerates any taint.

knelasevero: Thanks! Will take note of that

PR #561, opened 2025-05-02
OCPNODE-3192: enable the scheduler plugin if the feature gate DynamicResourceAllocation is enabled

tkashem:
Use case: if the cluster admin enables `TechPreview`, which has `DynamicResourceAllocation` feature gate enabled, it should automatically enable all of DRA, including the scheduler plugin `DynamicResources`.

We shouldn't have to force the admin to do the extra step:
```
oc patch --type merge -p '{"spec":{"profile": "HighNodeUtilization", "profileCustomizations" 
{"dynamicResourceAllocation": "Enabled"}}}' scheduler cluster
```


ingvagabund:
@tkashem let's wait for the branching. Unless this is needed in 4.19 as well.
/hold

tkashem:
> @tkashem let's wait for the branching. Unless this is needed in 4.19 as well.

+1 (it's not strictly required for 4.19)

tkashem:
/hold

(let's wait for CR to be little bit more stable)

PR #572, opened 2025-08-29
OCPBUGS-63213: Add priority field to prevent early shutdown

CoreyCook8:
Based on the issue described here : https://github.com/kubernetes/kubernetes/issues/133442

priorityClassName is currently ignored by Kubelet for static pod files so setting this value has no impact on the gracefulShutdown order causing the static pods to start to be killed as soon as shutdown begins.

To prevent this we must set priority explicitly

### DIFF
--- a/pkg/operator/targetconfigcontroller/testdata/ks_pod_scenario_3.yaml
+++ b/pkg/operator/targetconfigcontroller/testdata/ks_pod_scenario_3.yaml
@@ -148,4 +148,5 @@ spec:
   tolerations:
     - operator: Exists
   priorityClassName: system-node-critical
+  priority: 2000001000
 status: {}


CoreyCook8:
👋 @ingvagabund you're the recommended reviewer here, would you be able to take a look at these? 🙏 We opened 4 PRs to fix this issue, [one](https://github.com/openshift/cluster-etcd-operator/pull/1476) has been merged. I linked the other two below:

https://github.com/openshift/cluster-kube-controller-manager-operator/pull/865
https://github.com/openshift/cluster-kube-apiserver-operator/pull/1915

bindata/assets/kube-scheduler/pod.yaml
```diff
@@ -121,6 +121,7 @@ spec:
       - mountPath: /etc/kubernetes/static-pod-certs
         name: cert-dir
   hostNetwork: true
+  priority: 2000001000
```
ingvagabund, line 124: [Priority admission plugin](https://github.com/kubernetes/kubernetes/blob/c10a2932d483bea7bd8fc3b20b40995cae14f4f5/plugin/pkg/admission/priority/admission.go#L188-L190) mentions:
```
// if the pod contained a priority that differs from the one computed from the priority class, error
```
As long as the priority of `system-node-critical` PC does not change we are ok.

CoreyCook8: 👋 I created https://github.com/openshift/origin/pull/30268 if you don't mind taking a look and giving an ok-to-test?

PR #580, opened 2025-10-07
CNTRLPLANE-1544: Enable user namespace for the operator

tchap:
The operator now uses `hostUsers: false` in the associated deployment.
All relevant user and group IDs are set to 1000.


tchap:
/verified by @tchap 

When there is something wrong, CI typically breaks completely.

PR #588, opened 2025-11-12
OCPBUGS-65503: manifests: Expand related objects in clusteroperator

tchap:
Add system:openshift:operator:kube-scheduler-operator ClusterRoleBinding to the list of related objects.

### DIFF
--- a/manifests/0000_25_kube-scheduler-operator_07_clusteroperator.yaml
+++ b/manifests/0000_25_kube-scheduler-operator_07_clusteroperator.yaml
@@ -36,3 +36,6 @@ status:
     - group: controlplane.operator.openshift.io
       resource: podnetworkconnectivitychecks
       namespace: openshift-kube-scheduler
+    - group: "rbac.authorization.k8s.io"
+      name: "system:openshift:operator:cluster-kube-scheduler-operator"
+      resource: "clusterrolebindings"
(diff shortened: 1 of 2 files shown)


manifests/0000_25_kube-scheduler-operator_07_clusteroperator.yaml
```diff
@@ -36,3 +36,6 @@ status:
     - group: controlplane.operator.openshift.io
       resource: podnetworkconnectivitychecks
       namespace: openshift-kube-scheduler
+    - group: "rbac.authorization.k8s.io"
+      name: "system:openshift:operator:cluster-kube-scheduler-operator"
+      resource: "clusterrolebindings"
```
wking: The CVO will only apply these if it sees this manifest when the ClusterOperator doesn't exist.  That means these won't get injected into existing clusters updating into this new code.  Do you not reconcile these as part of your Go controller's ClusterOperator `status` management?  Or does this manifest somehow also feed your Go controller?  Or...?

tchap: You are right, I will mark this as WIP and research it a bit more.

tchap:
/verified by CI

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-kube-scheduler-operator/588/pull-ci-openshift-cluster-kube-scheduler-operator-main-e2e-aws-ovn-upgrade/1991779358502555648/artifacts/e2e-aws-ovn-upgrade/gather-extra/artifacts/inspect/cluster-scoped-resources/config.openshift.io/clusteroperators/kube-scheduler.yaml

PR #603, opened 2026-01-22
[release 4.21] OCPBUGS-74265: Add OPERATOR_IMAGE_VERSION to recovery container

tchap:
Set `OPERATOR_IMAGE_VERSION` environment variable on the recovery container in the operand pod so that the pod is redeployed during upgrade and we can see Progressing=True.

This does not make too much sense, but it's required for the operator to show `Progressing=True` during upgrade, and this is the way to achieve this. Other operators like KCMO work in a similar way, although there the value is actually being used.


tchap:
/verified bypass

This is basically tested by CI, but to confirm we will check the test statistics on Sippy.

PR #655, opened 2026-07-08
OCPBUGS-38663: Add 10-minute degraded inertia for StaticPodsDegraded

tchap:
Static pods may temporarily report errors during upgrades while pods restart. Use a longer inertia to avoid flapping the ClusterOperator Degraded condition.

## Summary by CodeRabbit

* **Refactor**
  * Simplified how degraded-state timing is configured, making the logic easier to maintain without changing behavior.

* **Tests**
  * Added coverage for an additional degraded condition to confirm the expected timeout is applied.


tchap:
The current CI failure is unrelated.

/verified by CI

PR #658, opened 2026-07-21
OCPBUGS-99295: add required-scc annotation to operator deployment pod

ropatil010:
This PR fixes the location of the required-scc annotation in the periodic-gathering job pod template

Add openshift.io/required-scc: restricted-v2 annotation to the kube-scheduler-operator deployment pod template to satisfy the SCC pinning requirement and resolve the flaking sig-auth test.

Failure job: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-shiftstack-ci-release-5.0-techpreview-e2e-openstack-ovn-serial-techpreview/2078882776429367296

Fixing this will allow to drop the [exception](https://github.com/openshift/origin/blob/main/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go#L62) from the required-scc-annotation-checker monitor test

## Summary by CodeRabbit

* **Security**
  * Updated the operator deployment to require OpenShift’s `restricted-v2` security constraint for its pods.

### DIFF
--- a/manifests/0000_25_kube-scheduler-operator_06_deployment.yaml
+++ b/manifests/0000_25_kube-scheduler-operator_06_deployment.yaml
@@ -19,6 +19,7 @@ spec:
       name: openshift-kube-scheduler-operator
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: restricted-v2
       labels:
         app: openshift-kube-scheduler-operator
     spec:


liouk:
@ropatil010 the namespace in which this workload lives (openshift-kube-scheduler-operator) is a [runlevel-0 namespace](https://github.com/openshift/cluster-kube-scheduler-operator/blob/main/manifests/0000_25_kube-scheduler-operator_00_namespace.yaml#L12) – therefore SCC admission does not run for those workloads. This means that the required-scc annotation will have no effect on this workload. This means that the reason these tests were failing is not related to SCC pinning, as the workloads don’t go through SCC admission.

I've closed the ticket as not-a-bug, closing this PR as well.

/close

